PR #27, opened 2019-09-17
Shrink UnsafeBuffer and JsBuffer after oversized messages

knah:
Well, WebSocketWire creates a new buffer for each send anyways, so no shrinkage there. It's supported though.


rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/AbstractBuffer.kt
```diff
@@ -64,6 +64,14 @@ abstract class AbstractBuffer {
 
     abstract fun checkAvailable(moreSize: Int)
 
+    open fun reset() {
```
korifey: Could you please add docs to make clear the difference between these two methods

PR #28, opened 2019-09-17
.NET Core tests, .NET Fw 3.5 compatibility, WinAPI ProcessWatchdog

epeshk:


operasfantom:
> Could you please make a branch in JetBrains repo and push it to our TeamCity server to verify everything is ok?

Do we have possibility to test github's PR in Teamcity, don't we?

rd-net/Test.RdFramework/SocketWireTest.cs
```diff
@@ -277,12 +278,14 @@ public void TestDisconnect()
         SpinWait.SpinUntil(() => log.Count == 2);
         Assert.AreEqual(new List<int> {1, 2}, log);
         
+        (clientProtocol.Wire as SocketWire.Base)?.Socket.Shutdown(SocketShutdown.Both);
```
korifey: You can make CloseSocket method static internal and call it.

PR #33, opened 2019-09-24
Deconstruct methods for C# generated classes #26

### LINKED ISSUE #26: Deconstruct methods for C# generated classes
"C# does not offer built-in support for deconstructing non-tuple types. However, as the author of a class, a struct, or an interface, you can allow instances of the type to be deconstructed by implementing one or more Deconstruct methods."
See https://docs.microsoft.com/en-us/dotnet/csharp/deconstruct

It would be great to see this in rd-gen! Even if it's supported only in C#7 and higher versions.

operasfantom:
Resolved #26


rd-kt/rd-gen/src/main/kotlin/com/jetbrains/rd/generator/nova/cpp/Cpp17Generator.kt
```diff
@@ -263,7 +284,7 @@ open class Cpp17Generator(override val flowTransform: FlowTransform, val default
 
     fun IType.substitutedName(scope: Declaration, rawType: Boolean = false, omitNullability: Boolean = false): String = when (this) {
         is Enum -> sanitizedName(scope)
-        is Struct.Concrete -> sanitizedName(scope)
+//        is Struct.Concrete -> sanitizedName(scope)
```
korifey: why we put comment instead of nothing?

PR #42, opened 2019-11-07
RdReflection related changes

ulex:
Use FQN in ReflectionSerializers to avoid clashes by Name
ReflectionSerializers: use interface name in RPC to support different names for impl/interface
Intrinsic: support intrinsic for RdModel (mostly for sake of UProperty)

### DIFF
--- a/rd-net/RdFramework/RdId.cs
+++ b/rd-net/RdFramework/RdId.cs
@@ -24,6 +24,11 @@ namespace JetBrains.Rd
       return new RdId(id ?? Hash(type.Name));
     }
 
+    public static RdId DefineByFqn(Type type)
+    {
+      return new RdId(Hash(type.FullName));
+    }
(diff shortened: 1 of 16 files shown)


rd-net/RdFramework/RdId.cs
```diff
@@ -14,16 +14,21 @@ public struct RdId : IEquatable<RdId>
...
       return new RdId(id ?? Hash(typeof(T).Name));
     }
 
-    public static RdId Define(Type type, int? id = null)
+    public static RdId Define(Type type, long? id = null)
     {
       return new RdId(id ?? Hash(type.Name));
     }
 
+    public static RdId DefineByFqn(Type type)
```
korifey: Please add comments that this method mustn't be used anyhow except C#-C# reflection interaction.

PR #50, opened 2019-12-03
Open entities support

nicklyu:


korifey:
Should I merge it or wait until interfaces?

nicklyu:
I think it's better to wait, I will commit them tonight or tomorrow morning

rd-kt/rd-gen/src/main/kotlin/com/jetbrains/rd/generator/nova/Types.kt
```diff
@@ -201,7 +207,12 @@ abstract class Declaration(open val pointcut: BindableDeclaration?) : SettingsHo
     protected abstract val cl_name: String
 
     open val base : Declaration? = null
-    open val isAbstract : Boolean get() = false
+    open val implements: MutableList<Interface> = mutableListOf()
```
korifey: any reason why it's mutable?

rd-kt/rd-gen/src/main/kotlin/com/jetbrains/rd/generator/nova/kotlin/Kotlin11Generator.kt
```diff
@@ -933,10 +958,14 @@ open class Kotlin11Generator(
 
 
     protected open fun PrettyPrinter.baseClassTrait(decl: Declaration) {
+        if(isUnknown(decl)){
+            decl.implements.add(Interface("IUnknownInstance", decl.pointcut as Toplevel, emptyList()))
```
korifey: I'm not sure we want to create new  Interface with same name for each class with such configuration. What is the purpose?

nicklyu: Fixed
But in general I think, that generators can and should update interfaces list with language-specific interfaces to avoid mixing constructions like 
`p(, IPrintable)` and generation from interfaces list

PR #52, opened 2019-12-04
Support for enums with initialized values in C++

operasfantom:


korifey:
Please use `switch` instead of if-else

operasfantom:
> Please use `switch` instead of if-else

See #53

PR #64, opened 2020-02-02
Heartbeat implementation

operasfantom:


rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -394,9 +506,11 @@ protected virtual bool AcceptHandshake(Socket socket)
 
     public class Client : Base
     {
-      public Client(Lifetime lifetime, [NotNull] IScheduler scheduler, int port, string optId = null) : this(lifetime, scheduler, new IPEndPoint(IPAddress.Loopback, port), optId) {}
+      public Client(Lifetime lifetime, [NotNull] IScheduler scheduler, int port, string optId = null, TimeSpan? heartBeatInterval = null) :
```
korifey: No need to add this parameter into contructor, make is as separate property so one could create Wire with customized interval: new SocketWire.Client(...) {HeartbeatInterval = ...}

rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -286,6 +386,18 @@ private int ReceiveFromSocket(byte[] buffer, int offset, int size)
       private const int PkgHeaderLen = sizeof(int) /*pkgFullLen */ + sizeof(long) /*seqN*/;
       private readonly byte[] mySendPkgHeader = new byte[ PkgHeaderLen];
       private readonly byte[] myAckPkgHeader = new byte[ PkgHeaderLen]; //different threads
+
+      public readonly TimeSpan HeartBeatInterval;
+      private readonly IViewableProperty<Timer> myTimer = new ViewableProperty<Timer>();
```
korifey: You never advise it. Why is it a property?

rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -273,6 +324,55 @@ private void SendAck(long seqN)
...
+        {
+          Log.Verbose($"{Id}: Socket was disposed during PING");
+        }
+        catch (Exception e)
+        {
+          Log.Warn(e, $"{Id}: ${e.GetType()} raised during PING");
+        }
+      }
+
+      private void SendPing()
```
korifey: inline this method

rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -224,8 +260,23 @@ private int ReceiveFromPkgBuffer(byte[] buffer, int offset, int size)
             Int32 len = UnsafeReader.ReadInt32FromBytes(myPkgHeaderBuffer.Data);
             Int64 seqN = UnsafeReader.ReadInt64FromBytes(myPkgHeaderBuffer.Data, sizeof(Int32));
 
-            if (len == ACK_MSG_LEN) 
+            if (len == PING_LEN)
+            {
+              myPonger.SendAsync(Unit.Instance);
+              continue;
+            } 
+            if (len == PONG_LEN)
```
korifey: Why do we need PONG events at all? Why PINGs aren't enough?

operasfantom: There could be different heartbeat intervals on client and server sides, so it wouldn't be enough to react only on PINGs. Also what's about multiclient connection at all?

rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -387,7 +391,7 @@ private int ReceiveFromSocket(byte[] buffer, int offset, int size)
       private readonly byte[] mySendPkgHeader = new byte[ PkgHeaderLen];
       private readonly byte[] myAckPkgHeader = new byte[ PkgHeaderLen]; //different threads
 
-      public readonly TimeSpan HeartBeatInterval;
+      public TimeSpan HeartBeatInterval { get; set; } = TimeSpan.FromMilliseconds(500);
       private readonly IViewableProperty<Timer> myTimer = new ViewableProperty<Timer>();
```
korifey: still don't get why is it a IViewableProperty

rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -42,6 +49,7 @@ public abstract class Base : WireBase
       protected readonly IViewableProperty<Socket> SocketProvider = new ViewableProperty<Socket> ();
       
       public readonly IViewableProperty<bool> Connected = new ViewableProperty<bool> { Value = false };
+      public readonly IViewableProperty<bool> StablyConnected = new ViewableProperty<bool> { Value = false };
```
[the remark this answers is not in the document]
operasfantom: What is good naming for that?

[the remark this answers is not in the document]
korifey: HeartbeatAlive

rd-net/RdFramework/Impl/SocketWire.cs
operasfantom: Okay, let it be the same intervals. Suppose the case when we lose outgoing messages, but successfully accept incoming messages. How to determine whether the connection is stable or not without reactions on PINGs?

korifey: How can this happen? Could you explain the example?

operasfantom: It can be happen due to network congestion only in one direction or firewall or smth like that. Are we going to exclude these cases from consideration?

korifey: This may be a valid point even if it's a corner case. But I still think we can use only one type of message with fully asynchronous interaction. Just put two additional fields in this message: timestamp of last received message from counterpart and your timestamp.

rd-kt/rd-framework/src/jvmMain/kotlin/com/jetbrains/rd/framework/SocketWire.kt
```diff
@@ -176,11 +217,22 @@ class SocketWire {
...
+                        counterpartTimestamp = timestamp
 
-                    if (len == ack_msg_len)
-                        sendBuffer.acknowledge(seqn)
+                        continue
+                    }
+
+                    val seqn = stream.readInt64() ?: return -1
+                    if (len == ack_msg_len){
+                        sendBuffer.acknowledge(seqn)}
```
knah: Formatting

rd-net/Test.RdFramework/SocketProxy.cs
```diff
@@ -0,0 +1,194 @@
...
+      myClientToServerLifetime = new SequentialLifetimes(myLifetime).With(lifetimes => lifetimes.Next());
+
+      myLifetime.OnTermination(() =>
+      {
+        myPort = null;
+        
+        StopServerToClientMessaging();
+        StopClientToServerMessaging();
+
+        myProxyServer?.Close();
```
knah: Close(0) (see the other PR about netcore bugs)

rd-net/Test.RdFramework/SocketProxy.cs
```diff
@@ -0,0 +1,194 @@
...
+          myPort = ((IPEndPoint) myProxyClient.LocalEndpoint).Port;
+          myLogger.Verbose($"Proxies for server on port {myServerPort} and client on port {Port} created successfully");
+
+          var acceptedClient = await myProxyClient.AcceptTcpClientAsync();
+
+          SetSocketOptions(acceptedClient);
+
+          myLogger.Verbose($"New client connected on port {Port}");
+
+          Connect(myProxyServer, acceptedClient);
```
knah: Consider closing the client socket somewhere too

rd-net/Test.RdFramework/SocketProxyTest.cs
```diff
@@ -0,0 +1,114 @@
...
+          var serverLog = new List<int>();
+          var clientLog = new List<int>();
+
+          sp.Advise(lifetime, i => serverLog.Add(i));
+          cp.Advise(lifetime, i => clientLog.Add(i));
+
+          //Connection is established for now
+
+          sp.Fire(1);
+          Thread.Sleep(SocketWireTest.DefaultTimeout);
```
knah: Use spinlocks instead of thread sleep

rd-net/Test.RdFramework/SocketProxy.cs
operasfantom: This is **TcpClient** and not **Socket**

operasfantom:
Look through the changes, please

operasfantom:
It's ready to be merged

operasfantom:
Is it okay for now?

PR #66, opened 2020-02-04
Implement cancelable + Bindable calls on kotlin side

Iliya-usov:


rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/TaskInterfaces.kt
```diff
@@ -95,11 +92,15 @@ val <T> IRdTask<T>.isSucceeded : Boolean get() = result.valueOrNull is RdTaskRes
 val <T> IRdTask<T>.isCanceled : Boolean get() = result.valueOrNull is RdTaskResult.Cancelled
 val <T> IRdTask<T>.isFaulted : Boolean get() = result.valueOrNull is RdTaskResult.Fault
 
+interface ILifetimedRdCall<in TReq, out TRes> {
```
korifey: This interface introduce new complexity. Why can't you mix everything into IRdCall ?

rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/TaskInterfaces.kt
```diff
@@ -120,10 +121,12 @@ interface IRdEndpoint<TReq, TRes> {
...
+    fun set(cancellationAndRequestScheduler: IScheduler? = null, handler: (Lifetime, TReq) -> RdTask<TRes>)
 
     /**
      * Assigns a handler that executes the API synchronously.
      */
-    fun set(handler: (TReq) -> TRes) = set { _, req -> RdTask.fromResult(handler(req)) }
+    fun set(cancellationAndRequestScheduler: IScheduler? = null, handler: (TReq) -> TRes) = set { _, req -> RdTask.fromResult(handler(req)) }
 }
+
+interface ILifetimedRdCallWithEndpoint<TReq, TRes> : ILifetimedRdCall<TReq, TRes>, IRdEndpoint<TReq, TRes>
```
korifey: Once again. Why do we need such an interface? We can generate RdCall for symmetric generation

rd-kt/rd-gen/src/main/kotlin/com/jetbrains/rd/generator/nova/csharp/CSharp50Generator.kt
```diff
@@ -180,6 +180,11 @@ open class CSharp50Generator(
                     Sink -> "IRdEndpoint"
                     Both -> "RdCall"
                 }
+                is Member.Reactive.BindableTask -> when (actualFlow) {
+                    Source -> "ILifetimedRdCall"
+                    Sink -> "IRdEndpoint"
+                    Both -> "ILifetimedRdCallWithEndpoint"
```
korifey: just RdCall

rd-net/RdFramework/Tasks/Interfaces.cs
```diff
@@ -17,11 +17,18 @@ public interface IRdEndpoint<TReq, TRes>
     void Set(Func<Lifetime, TReq, RdTask<TRes>> handler, IScheduler cancellationAndRequestScheduler = null);
   }
 
-  public interface IRdCall<in TReq, TRes>
+  public interface ILifetimedRdCall<in TReq, TRes>
```
korifey: Remove it

rd-net/RdFramework/Tasks/Interfaces.cs
```diff
@@ -17,11 +17,18 @@ public interface IRdEndpoint<TReq, TRes>
...
+  }
+
+  public interface IRdCall<in TReq, TRes> : ILifetimedRdCall<TReq, TRes>
   {
     TRes Sync(TReq request, RpcTimeouts timeouts = null);
     IRdTask<TRes> Start(TReq request, IScheduler responseScheduler = null);
+  }
 
-    IRdTask<TRes> Start(Lifetime lifetime, TReq request, IScheduler responseScheduler = null);
+  public interface ILifetimedRdCallWithEndpoint<TReq, TRes> : ILifetimedRdCall<TReq, TRes>, IRdEndpoint<TReq, TRes>
```
korifey: Same here

rd-kt/rd-gen/src/main/kotlin/com/jetbrains/rd/generator/nova/kotlin/Kotlin11Generator.kt
```diff
@@ -124,6 +129,7 @@ open class Kotlin11Generator(
     @Suppress("REDUNDANT_ELSE_IN_WHEN")
     protected open val Member.Reactive.implSimpleName : String get () = when (this) {
         is Member.Reactive.Task -> "RdCall"
+        is Member.Reactive.BindableTask -> "RdCall"
```
korifey: Maybe we have no need in separate BindableTask then?

rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/TaskInterfaces.kt
Iliya-usov: I agree. But we wanted to prohibit the use of the "Start" method for Bindable Tasks without Lifetime. 
Optionally, we can remove (or at least deprecate) the `start` method without lifetime. Then we won't need separate interfaces. What do you think about that?

korifey: Maybe we can achieve this goal by throwing exception from Start() method if lifetime isn't passed and return value is bindable.

Iliya-usov: I don't think it's a better solution to throw an exception in runtime if we can disable calling the `start` method without lifetime at compile level.
Do we really need the `start` method without lifetime. I think the best solution is to remove (mark as Deprecated) this method.

PR #70, opened 2020-02-06
Fixed hang on Mac .NET Core and hang in .NET tests

epeshk:


rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -107,8 +107,8 @@ internal static void CloseSocket([CanBeNull] Socket socket)
         
         //on netcore you can't solely execute Close() - it will hang forever
         //sometimes on netcoreapp2.1 it could hang forever during <c>Accept()</c> on other thread: https://github.com/dotnet/corefx/issues/26034 
-        // ourStaticLog.CatchAndDrop(() => socket.Close(TimeoutMs)); 
-        ourStaticLog.CatchAndDrop(socket.Close);
+        //we use zero timeout here to avoid blocking mode with (possible infinite) SpinWait
```
ForNeVeR: Could you please describe the conditions under which socket.Close() would lead to infinite SpinWait?

knah: According to reference source, non-zero timeouts (infinite -1 or positive numbers) lead to the problematic code with hanging spin wait. The linked corefx issue gives mixed feedback on when it's fixed (netcore 3/net 5), but we have first-hand evidence for issues on netcore 3.
Additionally, the worst thing that Close(0) does is sending a connection reset, which seems to be fine for our purposes.

PR #72, opened 2020-02-11
Fix upperbound check in EnumUtil.parseFromFlags()

nerzhulart:
Fix upperbound check in EnumUtil.parseFromFlags(). It was failing when trying to deserialize combination of last value in enum with some other value (see the test)

### DIFF
--- a/rd-kt/rd-core/src/commonMain/kotlin/com/jetbrains/rd/util/EnumUtil.kt
+++ b/rd-kt/rd-core/src/commonMain/kotlin/com/jetbrains/rd/util/EnumUtil.kt
@@ -9,7 +9,7 @@ inline fun <reified T:Enum<T>> parseFromOrdinal(ordinal: Int) : T {
 
 inline fun <reified T:Enum<T>> parseFromFlags(flags: Int) : EnumSet<T> {
     enumValues<T>().let { values ->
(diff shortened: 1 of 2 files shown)


rd-kt/rd-core/src/commonMain/kotlin/com/jetbrains/rd/util/EnumUtil.kt
```diff
@@ -9,7 +9,7 @@ inline fun <reified T:Enum<T>> parseFromOrdinal(ordinal: Int) : T {
 
 inline fun <reified T:Enum<T>> parseFromFlags(flags: Int) : EnumSet<T> {
     enumValues<T>().let { values ->
-        require(flags in 0..(1 shl (values.size - 1))) {"'$flags' not in range of ${T::class.simpleName} enum set: [0..${(1 shl (values.size - 1))})"}
+        require(flags in 0..(1 shl (values.size))) {"'$flags' not in range of ${T::class.simpleName} enum set: [0..${(1 shl (values.size - 1))})"}
```
knah: `0 .. ((1 shl values.size) - 1)` - the required range for a 3-flag enum is 000 to 111. Also, please fix the error string too.

PR #73, opened 2020-02-17
Close sockets immediately

ulex:
Motivation: most of the SocketWire tests are completed not due a proper interruption, but after a some magic timeout (you can observe it in test execution time and logs)
These delays became noticeable in the new unit test runner.

### DIFF
--- a/rd-net/RdFramework/Impl/SocketWire.cs
+++ b/rd-net/RdFramework/Impl/SocketWire.cs
@@ -470,12 +470,9 @@ namespace JetBrains.Rd.Impl
 
     public class Server : Base
     {
-      public Server(Lifetime lifetime, [NotNull] IScheduler scheduler, [CanBeNull] IPEndPoint endPoint = null, string optId = null) : this(lifetime, scheduler, CreateServerSocket(endPoint), optId)
-      {}
-
-      internal Server(Lifetime lifetime, IScheduler scheduler, Socket serverSocket, string optId = null) : base("ServerSocket-"+(optId ?? "<noname>"), lifetime, scheduler)
+      public Server(Lifetime lifetime, [NotNull] IScheduler scheduler, [CanBeNull] IPEndPoint endPoint = null, string optId = null) : this(lifetime, scheduler, optId)
       {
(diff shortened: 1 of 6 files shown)


rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -376,14 +376,21 @@ protected void AddTerminationActions([NotNull] Thread receiverThread)
               Monitor.PulseAll(Lock);
             }
 
+            CloseServerSocket();
```
knah: Server socket is already closed on server lifetime termination on line 502 (old) / 513 (new).
However, due to LIFO termination order of lifetimes, AddTerminationAction (and Thread.Join) are executed before socket-closing action from CreateServerSocket.
A possibly better approach would be to move all thread-joining actions later in termination order by adding the to a child LifetimeDef added to wire lifetime early in wire lifecycle. It's up for discussion though.

ulex: Nice catch, thanks. Didn't expect to find CloseSocket there. Introducing a child lifetime seems like over engineering to me. What about a contract of ownership transfer when SocketWire is created? I updated the review.

rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -617,7 +598,7 @@ public ServerFactory(Lifetime lifetime, IScheduler scheduler, IPEndPoint endpoin
         IPEndPoint endpoint = null
       )
       {
-        var serverSocket = Server.CreateServerSocket(lifetime, endpoint);
+        var serverSocket = Server.CreateServerSocket(endpoint);
```
knah: ServerFactory's serverSocket lifetime should be greater than lifetimes of individual Server instances it creates. As this is written right now, ServerFactory's socket will be closed after the first Server disconnects, which is not the desired behavior.

ulex: It does not work as you described (lifetime is common), but I have pushed theoretical fix.

PR #76, opened 2020-03-04
Change test annotations and junit versions to junit5

nicklyu:


rd-kt/rd-core/src/jvmTest/kotlin/com/jetbrains/rd/util/test/cases/LifetimeTest.kt
```diff
@@ -3,7 +3,6 @@ package com.jetbrains.rd.util.test.cases
 import com.jetbrains.rd.util.lifetime.*
 import com.jetbrains.rd.util.test.framework.RdTestBase
 import com.jetbrains.rd.util.threading.SpinWait
-import org.junit.Test
```
knah: Should we consistently use `org.junit.jupiter.api.Test` instead of `kotlin.test.Test` maybe?

nicklyu: Well, as I see, we use junit only in jvmTest submodule
I replaced all kotlin.test usages in jvmTest

operasfantom:
_KotlinJVMPlugin_ and corresponging tests in _rd-gen_ and _rd-text_ remained unchanged. But they still have _junitVersion_ which equals to 5.6.0. Should we bother about that?

operasfantom:
Seems like migration was finished successfully

PR #79, opened 2020-03-20
Fix leaks in WiredRdCalls

ulex:


ulex:
@korifey It would be great if you could take a look. I'm not sure if I understood semantic of externalCancellation corretly and that exactly did you mean by writing "todo"

korifey:
> @korifey It would be great if you could take a look. I'm not sure if I understood semantic of externalCancellation corretly and that exactly did you mean by writing "todo"

Could you describe memory leak mechanis in all details?

ulex:
Sure. In both cases problem occur via outerLifetime parameter, which, for
call
without external cancellation lifetime, is equal to BindableLifetime.
First issue occurs when explicitly adding a closure to BindableLifetime with
WiredRdTask in it. That means that on each call it store tons of useless
information.
https://github.com/JetBrains/rd/pull/79/commits/ba42d114483fb3a1310099ef21c58e1d707619f7
)
The second one is less critical. externalCancellation lifetime definition is
created in Subscribe method as a child for outerLifetime. Since this
definition
never goes beyond the boundary of this method and terminates only in
special cases,
all others LifetimeDefinition remain in memory until BindableLifetime is
terminated.
https://github.com/JetBrains/rd/pull/79/commits/adc793432c6dd77f83529a3d1c6c895fe19f1080

Issue #86, opened 2020-03-30
heartBeatInterval is immutable on frontend


slavam2605:
`com.jetbrains.rd.framework.SocketWire.Base#heartBeatInterval` should be `var` instead of `val`

ForNeVeR:
Currently, it is `var` everywhere.

PR #87, opened 2020-03-30
fix issue #86 | suppress warnings

operasfantom:
Upon moving _heartBeatInterval_ to _IWire_, its type changed to _kotlin.time.Duration_ which marked with annotation _ExperimentalTime_.

### DIFF
--- a/rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/base/RdExtBase.kt
+++ b/rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/base/RdExtBase.kt
@@ -189,7 +189,7 @@ class ExtWire : IWire {
     override val heartbeatAlive
         get() = realWire.heartbeatAlive
 
-    @UseExperimental(ExperimentalTime::class)
+    @OptIn(ExperimentalTime::class)
     override var heartbeatInterval: Duration
         get() = realWire.heartbeatInterval
         set(duration) {
--- a/rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/base/RdExtBase.kt
+++ b/rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/base/RdExtBase.kt
@@ -15,8 +15,6 @@ import com.jetbrains.rd.util.reactive.whenTrue
 import com.jetbrains.rd.util.string.printToString
 import com.jetbrains.rd.util.threading.SynchronousScheduler
 import com.jetbrains.rd.util.trace
-import kotlin.time.Duration
-import kotlin.time.ExperimentalTime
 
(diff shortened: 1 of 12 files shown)


rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/base/RdExtBase.kt
```diff
@@ -175,8 +177,15 @@ class ExtWire : IWire {
         require(newContexts === realWire.contexts) { "Can't replace ProtocolContexts on ExtWire" }
     }
 
+    @Suppress("ArrayInDataClass")
     data class QueueItem(val id: RdId, val msgSize: Int, val payoad: ByteArray, val context: List<Pair<RdContext<Any>, Any?>>)
     override val connected: Property<Boolean> = Property(false)
+    override val heartbeatAlive = connected
```
knah: Why not pass this to realWire too?

rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/base/RdExtBase.kt
```diff
@@ -175,8 +177,15 @@ class ExtWire : IWire {
...
+
+    @UseExperimental(ExperimentalTime::class)
+    override var heartbeatInterval: Duration
+        get() = Duration.INFINITE
+        set(_) {}
```
knah: Silently ignoring the set value seems questionable - it might be better to throw NotSupported

operasfantom: I have been thinking and realized that there is a scenario when user sets this property via IWire before connection is establsished. As well as it advises on heartbeatAlive.

ForNeVeR (approved):
Rebased on top of `master` and approved in principle.

ForNeVeR:
I have a couple of concerns, though.
1. This breaks our API with rename `heartBeatInterval` → `heartbeatInterval`.

   While I agree it is more grammatically correct (thus, better), I'm not sure it's worth to break the API right here.
2. I am a bit concerned about using the Kotlin experimental time API: what if it'll break in 1.5 or something like that? Say, our migration to 1.4 was painful partially due to "experimental" unsigned integers usage in our API. I would propose to get rid of `ExperimentalTime`, though I won't insist on it if everyone else is fine with the changes.

@mfilippov, your thoughts?

mfilippov:
1. Heartbeat is the grammatically correct word.
2. Totally agree with you. We shouldn't use the experimental API if it is possible.

ForNeVeR:
Decided to live with rename, but remove the experimental API. Alright.

mfilippov (approved):
LGTM

PR #110, opened 2020-06-11
Drop .NET Framework 4.6.1 support

mfilippov:


rd-net/Lifetimes/Lifetimes.csproj
```diff
@@ -46,6 +46,10 @@
     </PropertyGroup>
 
     <ItemGroup>
+        <PackageReference Include="JetBrains.NETFramework.ReferenceAssemblies.net35" Version="1.0.1">
```
ForNeVeR, line 49: Should we conditionally include it only for `net35`?

mfilippov: No, because this package has condition inside.

PR #113, opened 2020-06-16
Build rd cpp, pack to nuget and publish to GH packages

DecoyRS:


.github/workflows/build-rd-cpp.yml
```diff
@@ -0,0 +1,234 @@
...
+        path:  |
+          ${{ env.working-dir }}${{ matrix.config.SEP }}export${{ matrix.config.SEP }}**
+        name: ${{ env.artifact }}
+
+    - name: Pack Nuget
+      working-directory: ${{ env.working-dir }}${{ matrix.config.SEP }}export
+      run: nuget pack rd-cpp.nuspec -Version ${{ env.version }}
+
+    - name: Push nuget
+      uses: tanaka-takayoshi/nuget-publish-to-github-packages-action@v2.1
```
ForNeVeR, line 229: Currently, we'll push a package even for pull request, which shouldn't be the case. I suggest to change it as this:
```suggestion
      uses: tanaka-takayoshi/nuget-publish-to-github-packages-action@v2.1
      if: github.event_name == 'commit'
```

This should filter it to publish only the commits made on the master branch, and disable for pull requests.

.github/workflows/build-rd-cpp.yml
```diff
@@ -0,0 +1,234 @@
...
+            SEP: \
+          }
+        - {
+            name: "RD CPP Windows VS 2017", artifact: "RD-CPP-Windows-MSVC-2017-",
+            os: windows-2016,
+            cc: "cl", cxx: "cl",
+            environment_script: "C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/VC/Auxiliary/Build/vcvars64.bat",
+            SEP: \
+          }
+        - {
```
ForNeVeR, line 40: I suggest to disable clang, and only merge these two configurations once they become green.

.github/workflows/build-rd-cpp.yml
```diff
@@ -0,0 +1,239 @@
...
+        source-url: https://nuget.pkg.github.com/jetbrains/index.json
+      env:
+        NUGET_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
+    - name: Cache nuget
+      uses: actions/cache@v1
+      id: cache-nuget
+      with:
+        path: ~/.nuget/packages
+        key: ${{ runner.os }}-nuget-${{ hashFiles('**/packages.lock.json') }}
```
ForNeVeR, line 85: Why exactly do we need this? Are we installing any NuGet packages during the build? I thought we only use `nuget pack` and such. If so, then this step may be deleted.

Also, this seems to rely on `pacakages.lock.json` and we definitely aren't using these (yet).

rd-cpp/export/rd-cpp.nuspec
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata>
+        <id>JetBrains.RdFramework.Cpp</id>
+        <authors>JetBrains</authors>
+        <version>0.0.0.0</version>
+        <repository type="git" url="https://github.com/JetBrains/rd" />
+        <description>JetBrains Rd Framework CPP implementation</description>
+        <requireLicenseAcceptance>true</requireLicenseAcceptance>
```
ForNeVeR, line 9: While this shouldn't matter much, it's better to set this to `false`. `requireLicenseAcceptance=true` is usually reserved for cases when some unusual EULA is used. In our case, it's the standard Apache license.

But again, this doesn't matter much, because our main use cases don't include installing these packages via the standard NuGet client.

rd-cpp/src/rd_core_cpp/src/main/std/hash.cpp
```diff
@@ -0,0 +1 @@
+#include "hash.h"
```
ForNeVeR, line 1: I'm sorry, but why do we need `hash.cpp`, `unordered_map.cpp` and `unordered_set.cpp`? I'm not a C++ expert, but aren't template classes usually header-only? What't the use of these `.cpp`? They generate no executable code by themselves, right?

.github/workflows/build-rd-cpp.yml
DecoyRS: I've copy-pasted it from the example for caching nuget as I don't know anything about the way it's being stored. I'm ok with deleting it)

rd-cpp/src/rd_core_cpp/src/main/std/hash.cpp
DecoyRS: There were some gimmick with Ninja build that doesn't like it when there are solo headers.

mfilippov:
@DecoyRS  Could you squash it in a single commit?

PR #126, opened 2020-07-16
 Fix race in Serializers

ulex:
Current synchronization method has several drawbacks:
1. It does not work at all in NET3.5
2. Only actors implementation details protect against simultaneous write
3. It may lead to thread pool starvation

Because types can be registered in `ITypesRegistrar.TryRegister` method
during Read invocation, all others Read invocation on
ThreadScheduler.Default are forced to wait until Actor's queue is empty.

When many Reads are using TryRegister on many threads it may leads to
significant slow down during SpinWait on another threads which leads to
thread starvation sometimes (which is already happened on TeamCity)

### DIFF
--- a/rd-net/RdFramework/Impl/Serializers.cs
+++ b/rd-net/RdFramework/Impl/Serializers.cs
@@ -249,7 +249,9 @@ namespace JetBrains.Rd.Impl
         {
           if (myReaders.ContainsKey(typeId))
           {
-            var existingType = myTypeMapping.First(p => p.Value == typeId).Key;
+            Type existingType;
+            lock(myLock)
+              existingType = myTypeMapping.First(p => p.Value == typeId).Key;
             throw new ArgumentException(string.Format("Can't register {0} with id {1}. Already registered {2}", typeof(T).FullName, typeId, existingType));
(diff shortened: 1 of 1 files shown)


rd-net/RdFramework/Impl/Serializers.cs
```diff
@@ -277,14 +277,13 @@ public T Read<T>(SerializationCtx ctx, UnsafeReader reader, [CanBeNull] CtxReadD
...
+      if (!TryGetReader(typeId, out var ctxReadDelegate))
       {
         if (unknownInstanceReader == null)
         {
           myRegistrar?.TryRegister(typeId, this);
           myRegistrar?.TryRegister(typeof(T), this);
-          if (!myReaders.TryGetValue(typeId, out ctxReadDelegate))
+          if (!TryGetReader(typeId, out ctxReadDelegate))
           {
             var realType = myTypeMapping.SingleOrDefault(c => Equals(c.Value, typeId)); //ok because it's rarely needed
```
knah, line 288: myTypeMapping is read without a lock.
Additionally, myWriter is read without a lock on line 341

PR #127, opened 2020-07-22
Fix some compilation warnings and spamming runtime logged warnings in rd-cpp

LeoMash:


rd-cpp/src/rd_framework_cpp/src/main/scheduler/SingleThreadScheduler.cpp
```diff
@@ -16,6 +16,7 @@ SingleThreadScheduler::SingleThreadScheduler(Lifetime lifetime, std::string name
 		}
 		catch (std::exception const& e)
 		{
+			(void)e;
```
yvvan, line 19: You can instead just write `catch (std::exception const&)`
Same in next similar cases.

LeoMash: That's true, but I've intentionally left exception value named and have explicitly suppressed unused variable warning to be able to set a breakpoint inside catch block and inspect exception value without recompilation. I don't know if it's possible with anonymous catch block.

yvvan: Ah, I see. I guess that it gets eliminated by compiler in release mode anyways.

rd-cpp/CMakeLists.txt
```diff
@@ -29,7 +29,10 @@ if (CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj")
     endif ()
 elseif (MSVC)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
+    add_definitions(-D_CRT_SECURE_NO_WARNINGS)
+    add_compile_options(/wd4996     # POSIX name is deprecated
+                        /wd4250)    # method inheritance via dominance
```
DecoyRS, line 35: Same goes for compile options, would be better to have `target_compile_options` specified for problematic targets and avoid disabling warnings project wide

rd-cpp/CMakeLists.txt
```diff
@@ -29,7 +29,10 @@ if (CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj")
     endif ()
 elseif (MSVC)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
+    add_definitions(-D_CRT_SECURE_NO_WARNINGS)
```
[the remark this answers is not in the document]
LeoMash: My thoughts were that if rd is cross-platform library, then we still can't use "safe" msvc methods anywhere, so why bothering at all. Most of the crossplatform code I saw has suppressed _CRT_SECURE_NO_WARNINGS & wd4996 globally on whole project

PR #128, opened 2020-07-27
UnsafeWriter: option to clean thread static ourWriter after execution…

sergeyQx:
…. Fixes memory leak on heartbeat timer's callback that calls to Ping. Reduces memory leaks in RSRP-480566 Native memory leaked in UnsafeWriter

### DIFF
--- a/rd-net/Lifetimes/Serialization/UnsafeWriter.cs
+++ b/rd-net/Lifetimes/Serialization/UnsafeWriter.cs
@@ -24,13 +24,15 @@ namespace JetBrains.Serialization
     public struct Cookie : IDisposable
     {      
       private readonly UnsafeWriter myWriter;
+      private readonly Action myCleanupAction;
 
       private readonly int myStart;
 
       [MethodImpl(MethodImplAdvancedOptions.AggressiveInlining)]
-      public Cookie(UnsafeWriter writer) : this()
+      public Cookie(UnsafeWriter writer, Action cleanupAction = null) : this()
       {
         myWriter = writer;
+        myCleanupAction = cleanupAction;
         myStart = myWriter.Count;
       }
       
@@ -91,6 +93,7 @@ namespace JetBrains.Serialization
       public void Dispose()
       {
         myWriter?.Reset(myStart);
+        myCleanupAction?.Invoke();
       }
     }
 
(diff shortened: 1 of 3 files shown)


rd-net/Lifetimes/Serialization/UnsafeWriter.cs
```diff
@@ -24,13 +24,20 @@ public sealed unsafe class UnsafeWriter
...
       private readonly int myStart;
 
       [MethodImpl(MethodImplAdvancedOptions.AggressiveInlining)]
-      public Cookie(UnsafeWriter writer) : this()
+      public Cookie(UnsafeWriter writer) : this(writer, null)
+      {
+      }
+
+      [MethodImpl(MethodImplAdvancedOptions.AggressiveInlining)]
+      public Cookie(UnsafeWriter writer, Action cleanupAction = null) : this()
```
korifey, line 37: It's kind of dangerous to give user ability to invoke random cleanup actions after cookie is dispose. Do you really need such a power in public API?

rd-net/Lifetimes/Serialization/UnsafeWriter.cs
```diff
@@ -102,8 +110,29 @@ public void Dispose()
...
+    }
+
+    [MethodImpl(MethodImplAdvancedOptions.AggressiveInlining)]
+    public static Cookie NewThreadLocalWriterWithCleanup()
+    {
+      return NewThreadLocalWriterImpl(true);
+    }
+
+    [MethodImpl(MethodImplAdvancedOptions.AggressiveInlining)]
+    private static Cookie NewThreadLocalWriterImpl(bool releaseThreadStaticWriterIfNewlyCreated)
```
korifey, line 123: I guess the purpose of this method is to release 1MB of unsafe memory, when no cookies are on stack?

There are things to think about:

- Each time this method is used 1MB will be allocated and that is very time-consuming. Design of UnsafeWriter assumes that it's used on limited amount of threads with no penalty on creation new Cookie (except very first one). I guees this assumption was violated if you need this method. Maybe for short-living threads it's better to allocate less than 1Mb of memory?
- It's worth to deallocate unsafe memory instantly, not relying of finalizer ~UnsafeWriter then. Or you can play with weak handle to some object that holds 1Mb of memory, trying to pool it, not deallocating and allocating memory rapidly in case of frequest NewThreadLocalWriterWithCleanup() calls.

rd-net/Lifetimes/Serialization/UnsafeWriter.cs
sergeyQx: I do not see any reason for the Cookie to have public ctor in the first place. I do understand that it was created public t oavoid memory allocations

rd-net/Lifetimes/Serialization/UnsafeWriter.cs
sergeyQx: Unfortunately the heatbeat in SocketWire uses system timer and its callback is executed in random thread. The telemetry that Microsoft has shows that it leads to 50+ MB leaks quite frequently. 
It looks like there would be a better fix for it to create an UnsafeWriter over a preallocated buffer for the cases like sending the Ping message

rd-net/Lifetimes/Serialization/UnsafeWriter.cs
korifey: Yes, I think memory allocaton is the main reason.

rd-net/Lifetimes/Serialization/UnsafeWriter.cs
korifey: I see, thanks for explanation. Continuations on random threads and thread-bound memory do not work together well. The initial design assumed other usage and cookie creation should be as fast as possible. Maybe it worth to redesign memory allocation strategy for UnsageWriter, it's easy to abuse this API.
I can suggest special ThreadStatic variable (called e.g. "ShouldRetainMemory" in UnsafeWriter) that must be explicitly set for thread. Only if it's set UnsafeWriter takes memory from thread-local. Otherwiese - from pool.

sergeyQx: This is more or less the idea we agreed on while discussing internally (we used the term "our" threads) and will implement the fix by the ReSharper/ Rider 2020.3 RTM.  Or you can do it yourself if you wish))

korifey: I'd like so but It would be impossible for me to profile it on ReSharper and adjust my theoretical hypothesis according brutal reality. =)

PR #129, opened 2020-07-27
rd-cpp code cleanup: fix public dependencies and compilation warnings

LeoMash:


yvvan:
As I see you switch spdlog to non-headeronly version. Have you tried to build RiderLink with it? I had really bad link issues with it and that's why we use tweaked headeronly library.

LeoMash:
> As I see you switch spdlog to non-headeronly version. Have you tried to build RiderLink with it? I had really bad link issues with it and that's why we use tweaked headeronly library.

Yes, I've checked RiderLink with unity/non-unity x pch/nopch and it builds successfully. 
Header-only version exports windows.h header file that breaks compilation on some of these combinations, and it's unavoidable. So I hope that if I've missed something we will be able to fix it in library mode.

rd-kt/rd-gen/src/main/kotlin/com/jetbrains/rd/generator/nova/cpp/Cpp17Generator.kt
```diff
@@ -651,32 +652,43 @@ open class Cpp17Generator(flowTransform: FlowTransform,
 
     private fun File.cmakeLists(targetName: String, fileNames: List<String>, toplevelsDependencies: List<Toplevel> = emptyList(), subdirectories: List<String> = emptyList()) {
         mkdirs()
+        if (usingPrecompiledHeaders && !generatePrecompiledHeaders) {
```
DecoyRS, line 655: But why do we need to generate pch and don't use them?

LeoMash: I didn't change the way this option is used nowadays.
`usingPrecompiledHeaders` option describes if pch are enabled by default. But if this generated library is used as dependent project - outer project can override this value, but pch files should be generated and present in project anyway.
But we do not use cmake in ubt and we need extra option to not generate pch at all, that's why I added new option.

PR #130, opened 2020-07-28
Inline getAndUpdate method of AtomicReference as a potential workarou…

knah:
…nd for hangs near it, add more debug logs for possible causes of hangs

### DIFF
--- a/rd-kt/rd-core/src/commonMain/kotlin/com/jetbrains/rd/util/reactive/Signal.kt
+++ b/rd-kt/rd-core/src/commonMain/kotlin/com/jetbrains/rd/util/reactive/Signal.kt
@@ -14,7 +14,11 @@ class Signal<T> : ISignal<T> {
         fun priorityAdviseSection(block:() -> Unit) = incrementCookie(cookie, TlsBoxed<Int>::value) { block() }
 
         fun Void() = Signal<Unit>()
+
+        private val logger = getLogger<Signal<*>>()
     }
+
+    var debugId: String = "<no id>"
 
 
     private var priorityListeners = AtomicReference<Array<(T) -> Unit>>(emptyArray())
@@ -44,12 +48,20 @@ class Signal<T> : ISignal<T> {
                 {
                     queue.getAndUpdate { arr ->
                         if (arr.contains(handler)) throw IllegalArgumentException("Duplicate handler: $handler")
+                        if (arr.size == 10_000) {
+                            logger.warn { "Added over 10k handlers to a signal $debugId" }
+                        }
                         arr.insert(handler, arr.size)
                     }
                 },
                 {
-                    queue.getAndUpdate { arr ->
-                        arr.remove (handler).apply { if (equals(arr)) throw IllegalArgumentException("No handler: $handler") }
+                    val elapsed = measureTimeMillis {
+                        queue.getAndUpdate { arr ->
+                            arr.remove(handler).apply { if (equals(arr)) throw IllegalArgumentException("No handler: $handler") }
+                        }
+                    }
+                    if (elapsed > 1000) {
+                        logger.warn { "Removing a subscription from $debugId took $elapsed, have ${queue.get().size} subscriptions" }
                     }
                 }
         )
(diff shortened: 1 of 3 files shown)


rd-kt/rd-core/src/commonMain/kotlin/com/jetbrains/rd/util/reactive/Signal.kt
```diff
@@ -46,10 +48,18 @@ class Signal<T> : ISignal<T> {
                         if (arr.contains(handler)) throw IllegalArgumentException("Duplicate handler: $handler")
                         arr.insert(handler, arr.size)
                     }
+                    if (queue.get().size == 10_000) {
```
ForNeVeR, line 51: Wouldn't it be better to check this inside of `getAndUpdate` callback?
1) it won't require you to get the queue for the second time
2) I have a feeling that this code could easily _skip_ printing a message if there's high contention on an atomic (e.g. if two handlers were added between this and the previous step)

rd-kt/rd-core/src/commonMain/kotlin/com/jetbrains/rd/util/reactive/Signal.kt
```diff
@@ -46,10 +48,18 @@ class Signal<T> : ISignal<T> {
...
                 {
-                    queue.getAndUpdate { arr ->
-                        arr.remove (handler).apply { if (equals(arr)) throw IllegalArgumentException("No handler: $handler") }
+                    val elapsed = measureTimeMillis {
+                        queue.getAndUpdate { arr ->
+                            arr.remove(handler).apply { if (equals(arr)) throw IllegalArgumentException("No handler: $handler") }
+                        }
+                    }
+                    if (elapsed > 1000) {
+                        logger.warn { "Signal handler removal took longer than one second" }
```
ForNeVeR, line 62: I'm sorry, I'm not super familiar with our architecture, but could we print some useful information about the signal? E.g. its identifier? Smth like `model.foo.bar[0]`?

Otherwise, I'm not sure this warning would be super useful for us. We're almost sure that some signal handler removals take longer that one second; confirming this fact wouldn't do much for us.

knah: Unfortunately no. This signal is a local-only primitive (as opposed to RdSignal that is usually bound to a protocol tree), so no information is available. I'll remove this one then.

rd-kt/rd-core/src/jvmMain/kotlin/com/jetbrains/rd/util/PlatformDependent.kt
```diff
@@ -19,7 +19,15 @@ actual fun currentThreadName() : String = Thread.currentThread().run { "$id:$nam
...
-    actual fun getAndUpdate(f: (T) -> T): T = impl.getAndUpdate(f)
+    actual fun getAndUpdate(f: (T) -> T): T {
+        var prev: T
+        var next: T
+        do {
+            prev = impl.get()
+            next = f(prev)
+        } while (!impl.compareAndSet(prev, next))
+        return prev
+    }
```
ForNeVeR, line 30: Just as an explanation for interested reader (I know you're reading this!), we currently have a suspicion (unconfirmed yet), that in JDK11 this code could sometimes freeze for several minutes due to changes in `AtomicReference.getAndUpdate`, which now uses `weakCompareAndSet`. So, we're trying to roll it back to JDK8 behavior.

rd-kt/rd-core/src/commonMain/kotlin/com/jetbrains/rd/util/reactive/Signal.kt
ForNeVeR: I suggest to leave it here. Just as an additional confirmation of our main hypothesis, it will work

rd-kt/rd-core/src/commonMain/kotlin/com/jetbrains/rd/util/reactive/Signal.kt
knah: Fair point, especially the second one

rd-kt/rd-core/src/jvmMain/kotlin/com/jetbrains/rd/util/PlatformDependent.kt
korifey: 🤦‍♂️

PR #131, opened 2020-07-29
Improvement for RIDER-48322

ForNeVeR:
Roll back #130; log stack trace when 10k handlers are added.

It turns out that JDK wasn't the culprit of issue described in #130. After a dicussion, we've decided to roll these changes back, but keep the diagnostic about 10k handlers, and even call `log.error` in such a case, to print the stack trace to the log.

### DIFF
--- a/rd-kt/rd-core/src/commonMain/kotlin/com/jetbrains/rd/util/reactive/Signal.kt
+++ b/rd-kt/rd-core/src/commonMain/kotlin/com/jetbrains/rd/util/reactive/Signal.kt
@@ -18,8 +18,6 @@ class Signal<T> : ISignal<T> {
         private val logger = getLogger<Signal<*>>()
     }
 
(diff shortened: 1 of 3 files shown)


ForNeVeR:
@DecoyRS, take a look at the C++ build failure, please. It look like a publish problem. Should we publish pull request results at all?

DecoyRS:
> @DecoyRS, take a look at the C++ build failure, please. It look like a publish problem. Should we publish pull request results at all?

Oh no... Failing to publish nuget... Anything but this.:'(
Looking into it.

DecoyRS:
@ForNeVeR https://bintray.com/ is having a stroke at the moment.
You can check status here https://status.bintray.com/.
I wonder if we need to separate building and publishing to separate workflows and trigger release manually

ForNeVeR:
Alright, the CI has been fixed; I've rebased my commit.

PR #134, opened 2020-08-06
Qx unsafe writer

sergeyQx:
Refactored UnsafeWriter to cache the writer only on specifically marked threads. Whenr UnsafeWriter is used from a different thread its intial size reduced to 4KB. UnsafeWriterStatistics added.

### DIFF
--- a/rd-net/RdFramework/Impl/SocketWire.cs
+++ b/rd-net/RdFramework/Impl/SocketWire.cs
@@ -306,34 +306,31 @@ namespace JetBrains.Rd.Impl
 
       private void SendAck(long seqN)
       {
-        Scheduler.Queue(() =>
+        try
         {
-          try
+          using (var cookie = UnsafeWriter.NewThreadLocalWriterNoCaching())
           {
-            using (var cookie = UnsafeWriter.NewThreadLocalWriter())
-            {
-              cookie.Writer.Write(ACK_MSG_LEN);
-              cookie.Writer.Write(seqN);
-              cookie.CopyTo(myAckPkgHeader);
-            }
+            cookie.Writer.Write(ACK_MSG_LEN);
+            cookie.Writer.Write(seqN);
+            cookie.CopyTo(myAckPkgHeader);
+          }
 
-            lock (mySocketSendLock)
-              Socket.Send(myAckPkgHeader);
-          }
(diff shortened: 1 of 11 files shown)


rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -112,8 +112,8 @@ protected Base(string id, Lifetime lifetime, [NotNull] IScheduler scheduler) : b
       private Timer StartHeartbeat()
       {
         void OnTimedEvent(object sender, ElapsedEventArgs e)
-        {
-          Ping();
+        {
+          Scheduler.Queue(Ping);
```
knah, line 116: Using the default scheduler here may lead to wire being marked as potentially disconnected in cases of busy scheduler. With the old semantic pings were always sent regardless of what other threads were doing. Perhaps a ping scheduler should be added?

rd-net/Test.RdFramework/UnsafeWriterTest.cs
```diff
@@ -1,85 +1,176 @@
...
     public void TestWriterIsReused02()
     {
-      UnsafeWriter firstWriter;
-      using (var cookie = UnsafeWriter.NewThreadLocalWriter())
+      UnsafeWriter firstWriter = null, secondWriter = null;
+      var thread = new Thread(() =>
       {
-        firstWriter = cookie.Writer;
-      }
+        UnsafeWriter.AllowUnsafeWriterCaching = true;
```
knah, line 48: Given how widespread try-finally with this being set to true is, perhaps a cookie is in order?

rd-net/RdFramework/Impl/SocketWire.cs
sergeyQx: I just restored the behavior that existed before my changes. I introduced that Scheduler and then removed

rd-net/Test.RdFramework/UnsafeWriterTest.cs
sergeyQx: I do not expect many usages of the AllowUnsafeWriterCaching API. Most of the usages are in tests. Only rare threads are expected to allow cahcing. 
Cookie is possible

PR #138, opened 2020-08-20
change rd-gen dependency from org.jetbrains.kotlin:kotlin-compiler to…

van800:
… org.jetbrains.kotlin:kotlin-compiler-embeddable

Solves the following problem:
> rd-gen plugin depends on org.jetbrains.kotlin:kotlin-compiler:1.3.50. This is an incorrect way to depend on the Kotlin compiler, as it will bring the (unshaded) IJ platform classes as well as some other dependencies on the classpath
https://jetbrains.slack.com/archives/C0288G57R/p1597843597048000?thread_ts=1597761704.041400&cid=C0288G57R

### DIFF
--- a/rd-kt/rd-gen/build.gradle.kts
+++ b/rd-kt/rd-gen/build.gradle.kts
@@ -17,7 +17,7 @@ dependencies {
     compile(project(":rd-core:"))
     implementation(gradleApi())
     testImplementation(project(":rd-framework"))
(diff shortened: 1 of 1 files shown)


ForNeVeR (approved):

van800:
Ok, I have verified that new rd-gen allows setting
`id 'org.jetbrains.kotlin.jvm' version '1.4.0'`
in resharper-unity and locally everything works.

PR #141, opened 2020-08-27
Fix class loader issues in rdgen

ForNeVeR:


rd-net/Test.RdFramework/Reflection/data/Generated/RefRoot.cs
```diff
@@ -62,7 +62,6 @@ public static void RegisterDeclaredTypesSerializers(ISerializers serializers)
     {
       
       serializers.RegisterToplevelOnce(typeof(RefRoot), RefRoot.RegisterDeclaredTypesSerializers);
-      serializers.RegisterToplevelOnce(typeof(RefExt), RefExt.RegisterDeclaredTypesSerializers);
```
ForNeVeR, line 65: This change shouldn't be applied. It is, most likely, caused by a class loading bug in rdgen.

rd-kt/rd-gen/src/main/kotlin/com/jetbrains/rd/generator/nova/generate.kt
```diff
@@ -173,7 +195,11 @@ private fun collectSortedGeneratorsToInvoke(
 
     val gradle = gradleGenerationSpecs.map { it.toGeneratorAndRoot(roots) }
 
-    val external = classes.filter { IGeneratorAndRoot::class.java.isAssignableFrom(it) }.mapNotNull { it.kotlin.objectInstance }.filterIsInstance<ExternalGenerator>()
+    val foreignIGeneratorAndRoot = foreignClassLoader.loadClass(IGeneratorAndRoot::class.qualifiedName)
+    val generatorRootClasses = classes.filter { foreignIGeneratorAndRoot.isAssignableFrom(it) }
+    val currentClassLoader = IGeneratorAndRoot::class.java.classLoader
+    val reloadedGeneratorRootClasses = reloadInCompatibleClassLoader(currentClassLoader, generatorRootClasses)
+    val external = reloadedGeneratorRootClasses.mapNotNull { it.kotlin.objectInstance }.filterIsInstance<ExternalGenerator>()
```
ForNeVeR, line 202: I'm not sure about these changes. They were supposed to "reload" the classes in right class loader (like the ones in `collectTopLevels` — which work!), but seemingly they aren't.

On old (Kotlin 1.3) version, I'm getting the following report:
```
Collecting generators (filtering generator's `javaClass.simpleName` by regex '.*'):
  12 hardcoded generator(s) -- specified directly in 'Root' constructor
  0 gradle generator(s)    -- specified in gradle's rdgen plugin 'rdgen { generator { ...} }' sections
  0 external generator(s)  -- specified by extending kotlin object from class GenerationUnit
```

On the new one (Kotlin 1.4 with this patch applied):
```
Collecting generators (filtering generator's `javaClass.simpleName` by regex '.*'):
  29 hardcoded generator(s) -- specified directly in 'Root' constructor
  0 gradle generator(s)    -- specified in gradle's rdgen plugin 'rdgen { generator { ...} }' sections
  0 external generator(s)  -- specified by extending kotlin object from class GenerationUnit
```

This looks wrong and problematic.

ForNeVeR:
The underlying issue has been fixed by another means: 9318d8c1db679dccc8ccd35388b801804a0b98fe.

PR #142, opened 2020-08-27
Used Assembly.FullName instead of Assembly.GetName() to avoid failure…

epeshk:
… on systems without ICU package


mfilippov (approved):
LGTM

PR #145, opened 2020-08-30
Upload right test reports for failed tests

ForNeVeR:


ForNeVeR:
I've found that Gradle cache is broken, too. Will fix here.

ForNeVeR:
Tests are failing, but these're legit failures, not related to changes.

mfilippov (approved):
LGTM

PR #146, opened 2020-08-31
Bump Kotlin version to 1.4

mfilippov:


ForNeVeR (approved):
This is a set of crazy hacks to work around Kotlin 1.4 migration. But, as we can't see any other way without complete rewrite of build system, it has to be done this way for now.

We've scheduled complete rewrite of the most convoluted parts of the build system.

PR #147, opened 2020-09-03
Remove includes to thirdparty lib 'foonathan'

DecoyRS:
For some magical reasons, from time to time, some users log an error in the build where these includes will be evaluated, even though they are commented out.


DecoyRS:
These changes are stupid, yet I don't want any ninja commits, therefore I'm adding @mfilippov as a reviewer

ForNeVeR (approved):

ForNeVeR:
Please don't mind CI failure; it will be investigated separately.

PR #156, opened 2020-10-22
ToRdTask: nested tasks cancellation support

ulex:


ulex:
@knah sorry for messing up two separate commit into one PR, please avoid reviewing Reflection Serializers tests.

knah (approved):

PR #163, opened 2020-11-06
Fix package filter in rdgen

ForNeVeR:
Earlier, package filter in rdgen sometimes could capture the whole set of source classes passed to it, even it their packages weren't properly listed in the `packages` parameter.

From now on, it will not.


mfilippov (approved):
LGTM

PR #155, opened 2020-10-21
SequentialLifetimes: make sure SequentialLifetimes.IsCurrentTerminate…

sergeyQx:
Make sure SequentialLifetimes.IsCurrentTerminated while current lifetime termination

PR #165, opened 2020-11-10
SequentialLifetimes: thread-safe implementation, new termination behavior

ForNeVeR:
This updates the Kotlin implementation of `SequentialLifetimes` logic according to the .NET logic that was implemented in #155.


mfilippov (approved):
LGTM

PR #170, opened 2020-11-19
Update setup-dotnet to v1.7.2

ForNeVeR:
This is to avoid errors when calling `set-env` and `add-path`, see
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


mfilippov (approved):
LGTM

PR #172, opened 2020-11-25
LifetimeDefinition.Terminate: when termination fails to complete with…

sergeyQx:
…in certain timeout log WARNING (because it might not be our fault) and log ERROR when ExecuteIfAlive cookie is disposed

See https://jetbrains.slack.com/archives/CAZEUK2R0/p1606236742208100 for details


mfilippov (approved):
LGTM

PR #175, opened 2020-12-03
Disable unstable SocketProxyTest

ForNeVeR:
Later, I'll investigate what's wrong with it. For now, it does more wrong than good for the overall ecosystem.


mfilippov (approved):
LGTM

PR #179, opened 2020-12-09
Rd call scheduler usov

Iliya-usov:


rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/impl/RdTask.kt
```diff
@@ -228,22 +231,23 @@ class RdCall<TReq, TRes>(internal val requestSzr: ISerializer<TReq> = Polymorphi
...
      * Assigns a handler that executes the API synchronously.
      */
-    constructor(cancellationAndRequestScheduler: IScheduler? = null, handler: (TReq) -> TRes) : this () { set(cancellationAndRequestScheduler, handler) }
+    constructor(cancellationScheduler: IScheduler? = null, handlerScheduler: IScheduler? = null, handler: (TReq) -> TRes) : this () { set(cancellationScheduler, handlerScheduler, handler) }
 
     override fun onWireReceived(buffer: AbstractBuffer) {
         val taskId = RdId.read(buffer)
 
-        val wiredTask = EndpointWiredRdTask(bindLifetime, this, taskId, endpointSchedulerForHandlerAndCancellation ?: wireScheduler)
+        val wiredTask = EndpointWiredRdTask(bindLifetime, this, taskId, cancellationScheduler ?: SynchronousScheduler)
```
ForNeVeR, line 250: A couple of side notes:
- Yes, seems like `endpointSchedulerForHandlerAndCancellation` was only ever used for cancellation, and weren't actually used for _handler_, even if its name suggests otherwise. So, the corresponding API change is non-breakling one in practice.
- This is an (intended) behaviorial change: by default, `SynchronousScheduler` will be used for cancellation. We'll need to port it to C++ part, I guess.

rd-net/Test.Lifetimes/Collections/Viewable/DefaultSchedulerTest.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+          TestLogger.ExceptionLogger.Error(e);
+        }
+        finally
+        {
+          hasValue = true;
+        }
+      });
+
+      SpinWait.SpinUntil(() => hasValue, TimeSpan.FromSeconds(10));
```
ForNeVeR, line 42: Reading of variable set by other thread here is dubious at best. I'd recommend to use some sort of synchronization here: either `Volatile.Read/Write`, or just roll to `ManualResetEvent` instead of just `bool`.

rd-net/Test.Lifetimes/Collections/Viewable/SequentialSchedulerTest.cs
```diff
@@ -0,0 +1,167 @@
...
+          var reached = false;
+          scheduler.Queue(() =>
+          {
+            reached = true;
+            Assert.IsTrue(scheduler.IsActive);
+          });
+          
+          SpinWait.SpinUntil(() => reached, TimeSpan.FromMilliseconds(100));
+          
+          Assert.IsFalse(reached);
```
ForNeVeR, line 65: Here, I have the same comment about the `reached` variable as for the `hasValue` variable above. Probably it'd be better to avoid writing and reading an ordinary stack variable from different threads without any special measures applied.

Same for the `count` and `reached` variables below, in the next test.

rd-net/Test.Lifetimes/Collections/Viewable/DefaultSchedulerTest.cs
Iliya-usov: I don't think it's necessary in this case, but I used ManualResetEvent.

rd-net/Test.Lifetimes/Collections/Viewable/SequentialSchedulerTest.cs
Iliya-usov: SequentialScheduler guarantees memory consistency.

ForNeVeR (approved):

ForNeVeR (approved):

ForNeVeR (approved):

PR #190, opened 2020-12-21
Additional logging for errors in ByteBufferAsyncProcessor

ForNeVeR:
Since exceptions in this method may indicate a fatal error in the logging subsystem, the only place we're able to log them is `LogLog`.

### DIFF
--- a/rd-net/Lifetimes/Threading/ByteBufferAsyncProcessor.cs
+++ b/rd-net/Lifetimes/Threading/ByteBufferAsyncProcessor.cs
@@ -483,30 +483,38 @@ namespace JetBrains.Threading
 //          return;
 //        }
 
-        var ptr = 0;
-        while (ptr < count)
+        try
         {
-          Assertion.Assert(myChunkToFill.IsNotProcessed, "myChunkToFill.IsNotProcessed");
-          var rest = count - ptr;
-          var available = ChunkSize - myChunkToFill.Ptr;
-          if (available > 0)
+          var ptr = 0;
+          while (ptr < count)
           {
-            var copylen = Math.Min(rest, available);
-            Marshal.Copy((IntPtr)(start + ptr), myChunkToFill.Data, myChunkToFill.Ptr, copylen);
-            myChunkToFill.Ptr += copylen;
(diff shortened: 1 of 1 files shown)


rd-net/Lifetimes/Threading/ByteBufferAsyncProcessor.cs
```diff
@@ -483,30 +483,38 @@ public bool Terminate(int timeoutMs = Timeout.Infinite)
...
           }
         }
-
-        if (myAllDataProcessed) //speedup
+        catch (Exception ex)
         {
-          myAllDataProcessed = false;
-          Monitor.Pulse(myLock);
+          LogLog.Error(ex);
+          throw;
```
sergeyQx, line 517: (Note that rethrowing with “throw;” instead of “throw e;” would “preserve the corrupting-ness”, meaning the whole process would probably be torn down as the re-thrown exception continued to propagate.)

sergeyQx: @hypersw suggests to use throw new Exception(e) to preserve the original callstack

ForNeVeR: For this case, I don't think that's necessary.

Here, I want the following behavior from this code:
1. If any exception (including the corrupted state one) is thrown, then the lock should be unlocked, and the exception should be logged without reentrancy into this code region, if possible (`LogLog` allows such logging).
2. If the caller of this code uses `[HandleProcessCorruptedStateExceptions]`, then let him to handle the exception as it wants to.
3. If the caller of this code doesn't use `[HandleProcessCorruptedStateExceptions]`, then fall back to the default behavior (say, crash the whole process).

My point is that the logger internals shouldn't modify the defaults chosen by the application domain policy or by the caller. I don't think that this code should care about transforming corrupted state exceptions into ordinary ones.

ForNeVeR: After an internal discussion, we've decided to make this behavior configurable.

PR #197, opened 2021-02-04
Fix memory leak via ExtWire for stub wires

ulex:
Currently, every ext invocation is recorded and replayed when exts are
connected. For local protocol two exts will never be connected and every
wire packet will be recorded for ages.

To avoid storing these packets for stub wires special property is introduced.

### DIFF
--- a/rd-net/RdFramework/Base/RdExtBase.cs
+++ b/rd-net/RdFramework/Base/RdExtBase.cs
@@ -202,9 +202,12 @@ namespace JetBrains.Rd.Base
     
     public void Send<TContext>(RdId id, TContext param, Action<TContext, UnsafeWriter> writer)
     {
+      if (RealWire.IsStub)
+        return;
+
       lock (mySendQ)
       {
-        if (!RealWire.IsStub && (mySendQ.Count > 0 || !Connected.Value))
+        if (mySendQ.Count > 0 || !Connected.Value)
         {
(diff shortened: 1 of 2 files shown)


rd-net/RdFramework/Base/RdExtBase.cs
```diff
@@ -202,7 +204,7 @@ public void Send<TContext>(RdId id, TContext param, Action<TContext, UnsafeWrite
     {
       lock (mySendQ)
       {
-        if (mySendQ.Count > 0 || !Connected.Value)
+        if (!RealWire.IsStub && (mySendQ.Count > 0 || !Connected.Value))
```
ForNeVeR, line 207: For performance reasons, you may want to extract this new check outside of `lock`, to completely avoid locking on `mySendQ` for local connections.

ForNeVeR (approved):
Other than that, LGTM.

rd-net/RdFramework/Base/RdExtBase.cs
ulex: yep, it also does not make sense to call send of the real wire when it is stub. Fixed, thanks

ForNeVeR:
Oh gosh, Kotlin tests started being flaky again 😱

ulex:
Since this PR contains breaking API change, I will merge it later and update dotnet-products myself, after @mfilippov will overcome current issue with the C++/CLI compilation

PR #198, opened 2021-02-04
Allow delaying message processing until protocol is ready to handle them

knah:
This fixes context-caused exceptions

### DIFF
--- a/rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/MessageBroker.kt
+++ b/rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/MessageBroker.kt
@@ -14,10 +14,35 @@ class MessageBroker(private val defaultScheduler: IScheduler) : IPrintable {
         val log = Protocol.sublogger("MQ")
     }
 
+    constructor(defaultScheduler: IScheduler, queueMessages: Boolean) : this(defaultScheduler) {
+        isQueueingAllMessages = queueMessages
+    }
+
     private val lock = Any()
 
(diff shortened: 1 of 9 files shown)


rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/MessageBroker.kt
```diff
@@ -14,11 +14,36 @@ class MessageBroker(private val defaultScheduler: IScheduler) : IPrintable {
...
+
     private val lock = Any()
 
     private val subscriptions = hashMapOf<RdId, IRdWireable>()
     private val broker = hashMapOf<RdId, Mq>()
 
+    private var isQueueingAllMessages = false
+
+    fun startDeliveringMessages() {
+        require(isQueueingAllMessages) { "Already delivering messages" }
```
ForNeVeR, line 29: Shouldn't we move the check under the lock?

rd-net/RdFramework/Impl/MessageBroker.cs
```diff
@@ -74,8 +84,27 @@ private unsafe void Execute(IRdWireable reactive, byte[] msg)
             reactive.OnWireReceived(reader);
       }
     }
-    
-    
+
+    public void StartDeliveringMessages()
+    {
+      Assertion.Require(myIsQueueingAllMessages, "Already started delivering messages");
```
ForNeVeR, line 90: The same as above, here we're touching `myIsQueueingAllMessages` without a lock which is supposed to guard it. Should we move it inside of the `lock` section?

ForNeVeR (approved):
LGTM aside from a couple of little comments.

PR #116, opened 2020-06-25
Increments myTotalMessagesProcessed as an atomic operation

ulex:
It is a possible fix for ReSharper support case:
```
008fdaa8 7724216c [HelperMethodFrame: 008fdaa8] System.Threading.Thread.SleepInternal(Int32)
008fdb2c 71febe7b System.Threading.Thread.Sleep(Int32)
008fdb34 72084fc6 System.Threading.SpinWait.SpinOnce()
008fdb44 2479e89e JetBrains.Threading.SpinWaitEx.SpinUntil(JetBrains.Lifetimes.Lifetime, Int64, System.Func`1<Boolean>)
008fdb68 2479e801 JetBrains.Threading.SpinWaitEx.SpinUntil(System.Func`1<Boolean>)
008fdb88 2479e6fd JetBrains.Threading.Actor`1[[JetBrains.Rd.Impl.Serializers+ToplevelRegistration, JetBrains.RdFramework]].WaitForEmpty()
008fdb9c 2479e508 JetBrains.Rd.Impl.Serializers.Write[[System.__Canon, mscorlib]](JetBrains.Rd.SerializationCtx, JetBrains.Serialization.UnsafeWriter, System.__Canon)
008fdbd0 2479e4cc JetBrains.Rd.Impl.Polymorphic`1+<>c[[System.__Canon, mscorlib]].<.cctor>b__3_1(JetBrains.Rd.SerializationCtx, JetBrains.Serialization.UnsafeWriter, System.__Canon)
008fdbf0 2479e3eb JetBrains.Rd.Impl.RdProperty`1+<>c__DisplayClass21_0[[System.__Canon, mscorlib]].<Init>b__3(JetBrains.Rd.Impl.SendContext`2<System.__Canon,JetBrains.Rd.Impl.RdProperty`1<System.__Canon>>, JetBrains.Serialization.UnsafeWriter)
008fdc38 2479e15b JetBrains.Rd.Base.ExtWire.Send[[JetBrains.Rd.Impl.SendContext`2[[System.__Canon, mscorlib],[System.__Canon, mscorlib]], JetBrains.RdFramework]]
```

PR #210, opened 2021-04-12
Drop Actor from Serializers

ulex:
Motivated by:
[RSRP-483944](https://youtrack.jetbrains.com/issue/RSRP-483944) 2021.1 Beta: R# Unit Testing Take Nearly 30 Seconds to Start
https://github.com/JetBrains/rd/pull/119
https://github.com/JetBrains/rd/pull/116#issuecomment-650223785

### DIFF
--- a/rd-net/RdFramework/Impl/Serializers.cs
+++ b/rd-net/RdFramework/Impl/Serializers.cs
@@ -1,15 +1,12 @@
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
-using System.Linq.Expressions;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
 using JetBrains.Core;
 using JetBrains.Diagnostics;
 using JetBrains.Lifetimes;
 using JetBrains.Serialization;
-using JetBrains.Threading;
 using JetBrains.Util.Util;
 
 namespace JetBrains.Rd.Impl
@@ -35,41 +32,33 @@ namespace JetBrains.Rd.Impl
 
     [CanBeNull] private readonly ITypesRegistrar myRegistrar;
     private readonly object myLock = new object();
-
-    struct ToplevelRegistration
-    {
-      internal Type Type { get; }
-      internal Action<ISerializers> Action { get; }
-
-      public ToplevelRegistration(Type type, Action<ISerializers> action)
-      {
-        Type = type;
-        Action = action;
-      }
-    }
     
-  
 #if !NET35
-    private readonly Actor<ToplevelRegistration> myBackgroundRegistrar;
+    private readonly StealingScheduler myBackgroundRegistrar;
 
-    [Obsolete("Provide Lifetime and TaskScheduler. An active Task will leak in case of missing Lifetime (via Actor)", false)]
-    public Serializers() : this(Lifetime.Eternal, null, null)
+    public Serializers() : this(null, null)
     {
     }
 
-    public Serializers(Lifetime lifetime, [CanBeNull] TaskScheduler scheduler, [CanBeNull] ITypesRegistrar registrar)
+    public Serializers([CanBeNull] TaskScheduler scheduler, [CanBeNull] ITypesRegistrar registrar)
     {
       myRegistrar = registrar;
-      myBackgroundRegistrar = new Actor<ToplevelRegistration>("RegisterSerializers", lifetime, RegisterToplevelInternal, scheduler);
-      myBackgroundRegistrar.SendBlocking(new ToplevelRegistration(typeof(Serializers), RegisterFrameworkMarshallers));
+      myBackgroundRegistrar = new StealingScheduler(new ConcurrentExclusiveSchedulerPair(scheduler ?? TaskScheduler.Default).ExclusiveScheduler, false);
+      RegisterToplevelOnce(typeof(Serializers), RegisterFrameworkMarshallers);
     }
 
-    [Obsolete("Provide Lifetime and TaskScheduler. An active Task will leak in case of missing Lifetime (via Actor)", false)]
     public Serializers([CanBeNull] ITypesRegistrar registrar)
       : this()
     {
       myRegistrar = registrar;
     }
+
+    [Obsolete("Lifetime is not required anymore", false)]
+    public Serializers(Lifetime lifetime, [CanBeNull] TaskScheduler scheduler, [CanBeNull] ITypesRegistrar registrar)
+    : this(scheduler, registrar)
+    {
+    }
+
 #else
     public Serializers() => RegisterFrameworkMarshallers(this);
 
@@ -271,7 +260,7 @@ namespace JetBrains.Rd.Impl
           return myReaders.TryGetValue(rdId, out readDelegate);
       }
 #if !NET35
-      myBackgroundRegistrar.WaitForEmpty();
+      myBackgroundRegistrar.Join();
 #endif
       
       var typeId = RdId.Read(reader);
@@ -315,7 +304,7 @@ namespace JetBrains.Rd.Impl
       }
 
 #if !NET35
-      myBackgroundRegistrar.WaitForEmpty();
+      myBackgroundRegistrar.Join();
 #endif
 
       if (value == null)
@@ -351,21 +340,19 @@ namespace JetBrains.Rd.Impl
     private readonly HashSet<Type> myRegisteredToplevels = new HashSet<Type>();
     public void RegisterToplevelOnce(Type toplevelType, Action<ISerializers> registerDeclaredTypesSerializers)
     {
-      var r = new ToplevelRegistration(toplevelType, registerDeclaredTypesSerializers);
 #if !NET35
-      var task = myBackgroundRegistrar.SendOrExecuteInline(r);
-      Assertion.Assert(task.IsCompleted, "task.IsCompleted: {0}", task.Status);
+      new Task(() => RegisterToplevelInternal(toplevelType, registerDeclaredTypesSerializers)).Start(myBackgroundRegistrar);
 #else
-      RegisterToplevelInternal(r);
- #endif
+      RegisterToplevelInternal(toplevelType, registerDeclaredTypesSerializers);
+#endif
     }
 
-    private void RegisterToplevelInternal(ToplevelRegistration r)
+    private void RegisterToplevelInternal(Type type, Action<ISerializers> register)
     {
-      if (!myRegisteredToplevels.Add(r.Type)) return;
-      Protocol.InitTrace?.Log($"REGISTER serializers for {r.Type.Name}");
+      if (!myRegisteredToplevels.Add(type)) return;
+      Protocol.InitTrace?.Log($"REGISTER serializers for {type.Name}");
 
-      r.Action(this);
+      register(this);
     }
   }
 }
--- a/rd-net/RdFramework/Impl/StealingScheduler.cs
+++ b/rd-net/RdFramework/Impl/StealingScheduler.cs
@@ -12,12 +12,12 @@ namespace JetBrains.Rd.Impl
   /// </summary>
   public class StealingScheduler : TaskScheduler
   {
-    private readonly bool myAllowParallelProcessing;
+    private readonly bool myAllowParallelJoin;
     [NotNull] private readonly TaskScheduler myScheduler;
 
     /// <summary>
     /// >0: The number of currently simultaneously running tasks.
-    /// -1: exclusive mode taken by Join when <see cref="myAllowParallelProcessing"/> is false, no tasks allowed to process 
+    /// -1: exclusive mode taken by Join when <see cref="myAllowParallelJoin"/> is false, no tasks allowed to process 
     /// </summary>
     private volatile int myActive;
     [NotNull] private readonly ConcurrentQueue<Task> myActions = new ConcurrentQueue<Task>();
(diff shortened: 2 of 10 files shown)


korifey:
I guess you drop actor only from tests?
Historically, actor in Serializers is needed to reduce several seconds of JIT (to compile all delegates in all Exts) on application startup.

rd-net/RdFramework/Impl/Serializers.cs
```diff
@@ -348,24 +324,75 @@ bool TryGetTypeMapping(Type type1, out RdId rdId)
...
+          myQueue.Enqueue(new ToplevelRegistration(type, action));
+        }
+
+        var task = new Task(ProcessQueue);
+        task.Start(myScheduler);
+      }
+
+      public void ProcessQueue()
+      {
+        if (myQueueSize == 0)
```
ForNeVeR, line 378: Is it safe to read a non-volatile variable from outside of a lock? I'd recommend to declare `myQueueSize` as `volatile`.

rd-net/RdFramework/Impl/Serializers.cs
```diff
@@ -348,24 +324,75 @@ bool TryGetTypeMapping(Type type1, out RdId rdId)
...
+
+      public ToplevelRegistration(Type type, Action<ISerializers> action)
+      {
+        Type = type;
+        Action = action;
+      }
     }
 
-    private void RegisterToplevelInternal(ToplevelRegistration r)
+    private class BackgroundRegistrar
```
korifey, line 344: It's a very strange "fix". Maybe it worth to understand why actor doesn't work as expected (if it really happens) rather then reinventing the wheel and make yet another "background processer"?

ulex: > worth to understand why actor doesn't work as expected 

I've done this twice before. This is third time and have no idea how to fix it this time (and frankly speaking, have no desire to). IMO Rd.Actor is overcomplicated and I spent too much time figuring out that happened each time. 

Issue was caused by thread pool starvation and SendAsync from main thread. It is only a guess, since we cannot reproduce customer's issue on our side:
```
[TestCase(true)]
[TestCase(false)]
public void TestActorStarvation(bool withMainThread)
{
  var def = new LifetimeDefinition();
  var actor = new Actor<int>("TestActor", def.Lifetime, x => { });  
  if (withMainThread)
  {
    new Thread(() =>
    {
      Thread.Sleep(100);
      actor.SendBlocking(1);
    }) {IsBackground = true}.Start();
  }  for (int i = 0; i < 300; i++)
  {
    Task.Run(() =>
    {
      for (int j = 0; j < 10000; j++)
        actor.WaitForEmpty();
    });
  }  Task.Run(() => { }).Wait();
}
```

ulex:
> I guess you drop actor only from tests?

no, from the Serializer itself

rd-net/RdFramework/Impl/Serializers.cs
ulex: all primitive types up to 32 bits is safe to read outside of a lock (and get stale value sometimes, but it is not an issue here)

ForNeVeR: Only on x86, IIRC?

rd-net/RdFramework/Impl/Serializers.cs
korifey: Well, 
1) There is a Threal Pool in resharper that is defended against starvation. I don't remember exact name (maybe UnlimitedThreadPool) but it must be in JetBrains.Threading so you can move it here and make scheduler upon it.
BTW. It's easy to check starvation on .NET 45 - after 40 threads allocated each new thread requires 1 sec to initiate.
2) I don't know what do you mean about "overcomplicated" but reintventing wheel when you don't know exact reason of problem is not a good programming practice.

rd-net/RdFramework/Impl/Serializers.cs
ulex: I'm not aware of such restrictions and could not google it. I don't mind adding it, by the way (and I've added it).
I just wanted to understand in detail what exactly we are protecting from, and not adhere to some cargo-cult.

rd-net/RdFramework/Impl/Serializers.cs
sergeyQx: This is how ThreadManagementUnit works in ReSharper. But it also have issues - it is unable to detect thread starvation in some cases so I had to add a method that will tell it that the threads needs to be considered blocked.

@korifey in my understanding we do not need actors to reduce JIT time. We just need to process the serializers on background thread in some advance to thier actual use. And if we did not succeed just instantiate it sync when requested. This can be done in a controlled manner without scheduling the tasks on a thread pool hpoing for the best

rd-net/RdFramework/Impl/Serializers.cs
```diff
@@ -348,24 +324,75 @@ bool TryGetTypeMapping(Type type1, out RdId rdId)
...
+        {
+          if (!myRegisteredToplevels.Add(type))
+            return;
 
-      r.Action(this);
+          myQueueSize++;
+          myQueue.Enqueue(new ToplevelRegistration(type, action));
+        }
+
+        var task = new Task(ProcessQueue);
```
sergeyQx, line 372: This is a bit awkward to queue a task each time given the only one will do any job. But given this calls are seldom it it is ok

rd-net/RdFramework/Impl/Serializers.cs
ulex: @korifey 
1. You are proposing to overcome the caveats of actor's implementation by API consumers. Rd is used in many places nowadays: dotCover, Unit Testing, dotTrace, ReSharper OOP. Fixing all of them and tracking new misuses is beyond my pay grade.
2. "overcomplicated" is my personal opinion, as I mentioned. Any constructive feedback is appreciated.

rd-net/RdFramework/Impl/Serializers.cs
ulex: yep, but I prefer simplicity here over tiny performance gain

rd-net/RdFramework/Impl/Serializers.cs
korifey: @sergeyQx I got your point. As I mentioned before there is an UnlimitedThreadPool in ReSharper that handles starvation issue (you've even reviewed my code back there) and I propose to move it into Rd and use it everywhere in place of "handmade background processor". This pool try to use system thread pool and fall back to creating new threads only if starvation happens. 
UnlimitedThreadPool has attached TaskScheduler (per my memories, if not you can write one) that can be used by Actor to schedule activities sequentially on background.

@ulex I understand that you have to fit current business needs. Business continuity is a priority. On the other hand, one of  Rd's opensourcing reasons was to replace different historical R# implementation of one thing (like background processing) by canonical one (Actor).  Now what I see is a small step towards turning Rd back to "zoo" of reinvented wheels.
Actor relies havily on internal task scheduler so I propose just to change it to some scheduler without starvation. Or you know other issues in Actor?

ulex: The starvation itself was caused by synchronous await in Actor's code (waitForEmpy), see test I attached above. That I propose is to avoid this starvation cause, rather than treating the symptoms by throwing new threads. 
Although it may be worth to introduce some reusable thing.

korifey: So, if you propose to make some `waitAndPump` instead of `waitForEmpty`, it's a good idea that can be implemented directly in Actor. No need to make separate BackgroundRegistrar. It's so called "helper" approach when waiting thread actively stealing work from a queue. Fatal drawback here is that you invoke some potentially third-party code under stack frame of waiter thread. Usually we close our eyes on this drawback when mixing synchronous and asychronous code.

Few words attached test. I think that using `Task.Run()` is potentially flawed because of inevitable starvation on .NETFramework if something blocks inside. `WaitForEmpty()` here is just one example, but you will receive starvation every time on long blocking operation. Since it's hard to forbid blocking wait in general, I propose to use UnlimitedThreadPool everywhere in production to fix this problem once and for all.

rd-net/RdFramework/Impl/Serializers.cs
```diff
@@ -36,49 +34,30 @@ public class Serializers : ISerializers
...
     }
 
-    [Obsolete("Provide Lifetime and TaskScheduler. An active Task will leak in case of missing Lifetime (via Actor)", false)]
-    public Serializers([CanBeNull] ITypesRegistrar registrar)
-      : this()
+    public Serializers([CanBeNull] TaskScheduler scheduler, [CanBeNull] ITypesRegistrar registrar)
     {
       myRegistrar = registrar;
+      myBackgroundRegistrar = new BackgroundRegistrar(this, scheduler);
+      myBackgroundRegistrar.Register(typeof(Serializers), RegisterFrameworkMarshallers);
```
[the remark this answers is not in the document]
ulex: You are right, but I don't get the idea behind it and did not find any meaningful comment. I'm ready to return it back and write comment, if there is reason to such behavior

rd-net/RdFramework/Impl/Serializers.cs
```diff
@@ -348,24 +324,75 @@ bool TryGetTypeMapping(Type type1, out RdId rdId)
...
+
+      private volatile int myQueueSize;
+
+      public BackgroundRegistrar([NotNull] Serializers owner, [CanBeNull] TaskScheduler scheduler)
+      {
+        myOwner = owner;
+        myScheduler = scheduler ?? TaskScheduler.Default;
+      }
+
+      public void Queue(Type type, Action<ISerializers> action)
```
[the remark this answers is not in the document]
ulex: Yes, thanks! I've fixed in a updated version (StealingScheduler)

rd-net/RdFramework/Impl/Serializers.cs
ulex: > it's a good idea that can be implemented directly in Actor

Could you briefly describe how you see it? Processor isn't thread safe and there is no api to steal a task from an actor, so I thought it would be impossible without tremendeous hacks

Meanwhile I've updated PR and introduce StealingScheduler to make code cleaner and reusable if necessary.

> Since it's hard to forbid blocking wait in general, I propose to use UnlimitedThreadPool 

Thanks for clarifying. blocking can be avoided in this particular scenario and I believe it is worth it. And I have no control over the code that uses `Task.Run`, the attached test is only a model based on client side snapshot.

rd-net/RdFramework/Impl/StealingScheduler.cs
```diff
@@ -0,0 +1,134 @@
...
+
+    protected override IEnumerable<Task> GetScheduledTasks()
+    {
+      return myActions;
+    }
+
+    protected override void QueueTask(Task task)
+    {
+      myActions.Enqueue(task);
+      if (myActive < myScheduler.MaximumConcurrencyLevel)
```
ForNeVeR, line 47: Current implementation could lead to temporary bursts of activity when there are more `ProcessTasks` tasks scheduled than `myScheduler.MaximumConcurrencyLevel` (because `myActive` only considers threads that process a task at the current moment of time, and threads executing `ProcessTasks` could be in other states, too).

Say, if the scheduler is currently occupied by something else, then `myActive` will stay at zero, and thus any call to `QueueTask` will allocate and schedule a fresh task.

What do you think? Could/should we set a harder limit on the count of these tasks (say, try to terminate `ExecuteOne` in case of `myActive >= myScheduler.MaximumConcurrencyLevel`)? Or was it done deliberately the way it is?

ulex: The point of this check is not to strictly follow MaximumConcurrencyLevel of the wrapped scheduler, but rather just to avoid creating and queuing tasks on `myScheduler` when it is definitely pointless. We aware of the "proccess all" behavior of `StealingScheduler` and that is why it is valid to apply this small optimization.
Strictly enforcing the amount of concurrent threads is `myScheduler` responsibility.

ForNeVeR: I'm not talking about cases when several tasks run concurrently (since it isn't a production case even).

Consider a case when the scheduler is occupied by some long task, and there are several clients who want to do `QueueTask` (say, to register a new serializer). In this case, each client will cause the creation of a `new Task(ProcessTasks)`.

Later, after the scheduler is free, it will start processing these tasks (most likely, one by one), so the first task will do all the work, and others will do nothing, and will be GC'd later.

As I see the current implementation, it could create some excessive garbage on serializer registration, in cases when it isn't required.

Yes, it shouldn't be too much in practice (considering how we use this scheduler and how many serializers we register in each run), but still: don't you think we could do better here?

This is a genuine question; I'm not sure we should improve that, but I'd like to discuss.

Other than that, I have no concerns about this approach, but I'd like to wait for other reviewers. The scheduler code is tricky, and it's possible I've missed some important points.

Iliya-usov (approved):

rd-net/RdFramework/Impl/StealingScheduler.cs
ulex: Sorry for long answer, I've waited for a possible comments from other reviewers.

Thanks for clarification, I've understood the case, but cannot find a neat way to implement it without additional tracking of scheduled tasks. Also, the impact of such optimization is also debatable - task tracking isn't free.

In the current production code it seems as incredible rare scenario. We are using exclusive scheduler over default thread pool. This scheduler may be occupied for a long time when either thread pool starvation happens or exclusive scheduler already executing our task. In the second case we won't schedule new tasks because myActive == MaxConcurrencyLevel.

PR #213, opened 2021-04-27
SocketWireTest diagnostic improvements

ForNeVeR:
Current plan is to improve logging (done) and then continue investigation. 

Maybe the test only fail on IPv6 nodes, or something like that.


mfilippov (approved):
LGTM

PR #216, opened 2021-04-29
Fix mess in EOL

ulex:
Currently, most of the files in the repository have LF line-endings, but
there are 5 files with CRLF. This is because all the following
conditions have been met:
1. Visual Studio supports mixed line-endings
2. You can specify `core.autocrlf` to false or input on Windows to
   checkout files with the original line-endings.
3. `.editorconfig` was blindly copied from the dotnet-products
   repository. In this file eol for most of the files is set to CRLF.
4. When you edit an LF file in Visual Studio with an editorconfig file,
   that specifies CRLF for the same file, new lines will get CRLF, but
   others will remain LF [1]

[1]: https://developercommunity2.visualstudio.com/t/Visual-Studio-creates-different-line-end/1272220


ForNeVeR (approved):

ForNeVeR:
Test failure (flaky?) has been noted.

PR #217, opened 2021-04-29
Fixed default value in constructor generation.

vvselischev:


rd-kt/rd-gen/src/main/kotlin/com/jetbrains/rd/generator/nova/csharp/CSharp50Generator.kt
```diff
@@ -900,10 +900,17 @@ open class CSharp50Generator(
         indent {
             +decl.allMembers
                     .joinToString(",\n") {
-                        val defValue = getDefaultValue(decl, it) ?: ""
-                        if (!it.hasEmptyConstructor) sanitize(it.name)
-                        else "${it.creationExpressionSubstituted(decl)}(${(it as? Member.Reactive)?.customSerializers(decl, leadingComma = false)
-                                ?: ""}$defValue)"
+                        if (!it.hasEmptyConstructor) {
+                            sanitize(it.name)
```
ForNeVeR, line 904: Shouldn't we add `new` and `()` here?

vvselischev: No. In the section on lines 885-897 we print members without default constructor as arguments to the current secondary constructor. Here we simply pass the names of these arguments to the primary constructor.

ForNeVeR:
Could you add a test, please?

ForNeVeR (approved):

PR #219, opened 2021-05-05
Qx rd outer lifetime

sergeyQx:
RdOuterLifetime to pass OuterLifetime over the protocol


ulex:
I have no objection to merging it, but I think we should refrain from this scheme by default and use it only in special situations. There are existing implicit alternatives to Lifetime in protocol (like presense in live collections) which should be used in the most scenarios.

ulex (approved):
I have no objection to merging it, but I think we should refrain from this scheme by default and use it only in special situations. There are existing implicit alternatives to Lifetime in protocol (like presense in live collections) which should be used in the most scenarios.

korifey:
Totally agree with @ulex 

what is your scenario of passing lifetime explicitly? 
usually you set some property and view it to get lifetime on counterpart side

PR #220, opened 2021-05-05
Reflection: return RdModel with lifetime as parameter from Sync methods

ulex:


ulex:
@mfilippov FYI: I've updated LangVersion to 8.0 in this PR

ForNeVeR (approved):

PR #221, opened 2021-05-07
CI: increase the test timeout

ForNeVeR:
As of now, the tests may usually take up to 10–13 minutes, and with a timeout of 15 minutes, outliers may really fail the tests for no reason. I propose we increase the timeout for now.


mfilippov (approved):
LGTM

PR #226, opened 2021-06-09
Add special `valueForPerContextEntity` for RdContext 

nerzhulart:
Add special `valueForPerContextEntity` for RdContext  to use it inside RdPerContextMap. It allows to override the value for specific context to provide a default one and avoid throwing when accessing RdPerContextMap items

### DIFF
--- a/rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/RdContext.kt
+++ b/rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/RdContext.kt
@@ -52,6 +52,13 @@ abstract class RdContext<T : Any>(val key: String, val heavy: Boolean, val seria
         get() = internalValue.get()
         set(value) = internalValue.set(value)
 
+    /**
+     * Value which is used as a key inside per-context entities like [RdPerContextMap][com.jetbrains.rd.framework.impl.RdPerContextMap]
+     */
+    open var valueForPerContextEntity: T?
+        get() = value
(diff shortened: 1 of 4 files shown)


rd-kt/rd-framework/src/commonMain/kotlin/com/jetbrains/rd/framework/RdContext.kt
```diff
@@ -52,6 +52,13 @@ abstract class RdContext<T : Any>(val key: String, val heavy: Boolean, val seria
         get() = internalValue.get()
         set(value) = internalValue.set(value)
 
+    /**
+     * Value which is used as a key inside per-context entities like [RdPerContextMap][com.jetbrains.rd.framework.impl.RdPerContextMap]
+     */
+    open var valueForPerContextEntity: T?
+        get() = value
+        set(value) { this.value = value }
```
knah, line 60: Not sure if the setter is necessary here

rd-net/RdFramework/RdContext.cs
```diff
@@ -101,6 +101,15 @@ public T Value
       set => myValue.Value = value;
     }
 
+    /// <summary>
+    /// Value which is used as a key inside per-context entities like <see cref="RdPerContextMap{K,V}"/>
+    /// </summary>
+    public virtual T ValueForPerContextEntity
+    {
+      get => Value;
+      set => Value = value;
```
knah, line 110: Not sure if setter is necessary here

knah (approved):

PR #228, opened 2021-06-19
Fix a flaky rd-gen test

ForNeVeR:
1. This fixes the test failures [like this one](https://github.com/JetBrains/rd/runs/2865407895), when `com.jetbrains.rd.generator.test.cases.generator.example.TestExample` fails on Linux with an error "org.opentest4j.AssertionFailedError: Failed to compile generated sources: Compilation failed. Return code: 1(COMPILATION_ERROR)", with the following written to stderr:

   > error: unable to run REPL, no scripting plugin loaded

   The issue turns out to be about the generated Kotlin output directory.
   
2. For more clearer error messages about this in the future, I've added an explicit input file count check before Kotlin compiler invocation.
3. Linux tests are migrated to Docker, because without it I wasn't able to reliably reproduce the issue.
4. `RdGen` output is now `verbose` in all the rd-gen tests.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,16 @@
+**/.gradle/
+**/build/
+
+.github/
+.idea/
+rd-cpp/
+rd-net/
+
+.dockerignore
+.editorconfig
+.gitignore
+.gitmodules
+Dockerfile
+LICENSE
+NuGet.Config
+README.md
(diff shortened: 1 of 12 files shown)


mfilippov (approved):
LGTM

PR #245, opened 2021-09-08
Implementation of created extensions listener

slavam2605:


rd-kt/rd-framework/src/main/kotlin/com/jetbrains/rd/framework/base/RdExtBase.kt
```diff
@@ -51,8 +51,11 @@ abstract class RdExtBase : RdReactiveBase() {
         lifetime.bracket(
             {
 //                extScheduler = sc
-                extProtocol = Protocol(parentProtocol.name, parentProtocol.serializers, parentProtocol.identity, sc, extWire, lifetime, serializationContext, parentProtocol.contexts).also {
+                extProtocol = Protocol(parentProtocol.name, parentProtocol.serializers, parentProtocol.identity, sc, extWire, lifetime, serializationContext, parentProtocol.contexts, parentProtocol.extCreatedLocal, createExtSignal()).also {
                     it.outOfSyncModels.flowInto(lifetime, super.protocol.outOfSyncModels) { model -> model }
+                    parentProtocol.extCreatedNetworked.localChange {
```
kashmervil, line 56: Am I right that the code won't be executed for the case of local ext creation since the `lifetime` will be terminated?
And that is the guarantee behind that  
```
extCreatedNetworked.advise(lifetime) { message ->
  if (extCreatedNetworked.isLocalChange) return@advise
    extCreatedLocal.fire(message)
}
 ```
 
 If that's the case, I find it very confusing and hard to understand. It hides the real reasons by the peculiarities and corner cases of lifetime-related API that skips the execution for `Lifetime.Terminated`, while leaving out the code that uses local cookies for remote updates with unclear execution flow. It's hard to tell for sure if a particular API won't throw an exception in the case of `Lifetime.Terminated`, so usually people don't mess with terminated lifetimes. 
 
Sorry if that's not the case

**Review by kashmervil — CHANGES_REQUESTED** [review]

File: rd-kt/rd-framework/src/main/kotlin/com/jetbrains/rd/framework/base/RdBindableBase.kt [hunk 2]
```diff
@@ -67,7 +67,7 @@ abstract class RdBindableBase : IRdBindable, IPrintable {
                 bindLifetime = lf
             },
             {
-                bindLifetime = lf
+                bindLifetime = Lifetime.Terminated
```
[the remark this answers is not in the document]
knah: In previous version null would be more appropriate indeed, with other changes to this file it's approximately equivalent (but not keeping the reference to an old lifetime might be more memory-efficient)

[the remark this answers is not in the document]
slavam2605: I changed the JVM part to be consistent with the .NET part because the .NET part looks more correct.

rd-kt/rd-core/src/main/kotlin/com/jetbrains/rd/util/string/RName.kt
```diff
@@ -16,6 +16,21 @@ class RName private constructor(val parent:RName?, val localName: String, val se
      */
     fun sub(localName: String, separator: String) = RName(this, localName, separator)
 
+    fun getNonEmptyRoot(): RName {
```
ForNeVeR, line 19: An interesting consequence of this implementation is that `Empty.getNonEmptyRoot()` would be equal to `Empty`. I don't know how to solve this (or whether we even want to "solve" it), though.

rd-kt/rd-framework/src/main/kotlin/com/jetbrains/rd/framework/MarshallerUtils.kt
```diff
@@ -0,0 +1,58 @@
+package com.jetbrains.rd.framework
+
+import com.jetbrains.rd.framework.base.IRdWireable
+import com.jetbrains.rd.framework.impl.RdSignal
+import com.jetbrains.rd.util.string.RName
+
+object RNameMarshaller : IMarshaller<RName> {
```
ForNeVeR, line 7: Do we have any tests covering this?

rd-kt/rd-framework/src/main/kotlin/com/jetbrains/rd/framework/base/RdBindableBase.kt
```diff
@@ -26,8 +26,8 @@ abstract class RdBindableBase : IRdBindable, IPrintable {
 
     protected var parent : IRdDynamic? = null
 
-    private var bindLifetime: Lifetime? = null
-
+    lateinit var bindLifetime: Lifetime
```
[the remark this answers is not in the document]
ForNeVeR: Also, public `lateinit` looks extremely dangerous.

rd-net/RdFramework/Impl/MessageBroker.cs
```diff
@@ -228,6 +229,10 @@ public void Advise(Lifetime lifetime, IRdWireable reactive)
         });
     }
 
-
+    [CanBeNull]
+    public IRdWireable TryGetById(RdId rdId)
+    {
+      return mySubscriptions.TryGetValue(rdId, out var value) ? value : null;
```
ForNeVeR, line 235: Same as with the frontend side, `lock (myLock)` is required here.

ForNeVeR (changes requested):

rd-kt/rd-framework/src/main/kotlin/com/jetbrains/rd/framework/Interfaces.kt
```diff
@@ -38,6 +36,10 @@ interface IProtocol : IRdDynamic {
     val outOfSyncModels: ViewableSet<RdExtBase>
 
     val contexts : ProtocolContexts
+
+    val extCreatedNetworked: RdSignal<Triple<RName, RdId?, Long>>
```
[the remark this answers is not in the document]
slavam2605: `RdSignal` is important here because `localChange` is used. I don't think that cast it every time is better than having a more specific type in `IProtocol`. I can change a type of local signal to `ISignal` though.

rd-kt/rd-core/src/main/kotlin/com/jetbrains/rd/util/reactive/Scheduler.kt
```diff
@@ -32,4 +32,8 @@ interface IScheduler {
     }
 
     fun flush()
+}
+
+interface ISchedulerWithBackground : IScheduler {
```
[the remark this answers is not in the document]
slavam2605: Moved from `IRdDispatcher` with no documentation there. Could someone help me to come up with a correct description?

rd-kt/rd-framework/src/main/kotlin/com/jetbrains/rd/framework/base/RdExtBase.kt
slavam2605: This is a tricky way to use ext's delayed message sending. There is only one local signal, passed through all protocols, and a remote signal for each protocol instance (including ext protocols). It's quite complex, indeed, but has nothing to do with Lifetime API.

rd-kt/rd-core/src/main/kotlin/com/jetbrains/rd/util/string/RName.kt
slavam2605: Ideally, I would prefer to reimplement RName with a reversed list because the only case when straight order is needed is logging. But I don't want to change RName dramatically here, so I ended up with linear complexity implementations for every operation. It shouldn't be a performance problem since we don't have too long chains in RName.

rd-kt/rd-framework/src/main/kotlin/com/jetbrains/rd/framework/MarshallerUtils.kt
slavam2605: No. Good point, thanks.

kashmervil (changes requested):

kashmervil (approved):

ForNeVeR (approved):

knah (approved):

PR #249, opened 2021-09-15
Build: use SNAPSHOT version outside the TeamCity builds

ForNeVeR:
This way, `./gradlew publishToMavenLocal` will use the same `SNAPSHOT` directory by default (for local development) as earlier.


mfilippov (approved):
LGTM

PR #262, opened 2021-11-06
Fix lifetime leaks and improve socket test logging

ForNeVeR:


rd-kt/rd-framework/src/test/kotlin/com/jetbrains/rd/framework/test/util/TestBase.kt
```diff
@@ -35,6 +35,7 @@ open class TestBase {
     }
 
     @AfterEach
+    @Timeout(value = 30, unit = TimeUnit.SECONDS)
```
ForNeVeR, line 38: Why do we need this, isn't the whole `TestBase` already covered by the identical `@Timeout`?

Turns out: nope. As documented by `@Timeout`, test lifecycle methods (such as `@AfterEach`) aren't covered by the test-wide `@Timeout`. So, we'll have to add one here.

rd-kt/rd-framework/src/test/kotlin/com/jetbrains/rd/framework/test/cases/wire/SocketWireTest.kt
```diff
@@ -291,20 +291,20 @@ class SocketWireTest : TestBase() {
...
         val factory = SocketWire.ServerFactory(sLifetime, TestScheduler, 0)
 
-        val lf1 = LifetimeDefinition()
-        SocketWire.Client(lf1, TestScheduler, factory.localPort)
+        Lifetime.using { lf1 ->
+            SocketWire.Client(lf1, TestScheduler, factory.localPort)
 
-        spinUntil { factory.size == 1 }
-
-        val lf2 = LifetimeDefinition()
```
ForNeVeR, line 302: While everything else was more or less acceptable, `lf2` never was terminated.

Iliya-usov (approved):

Issue #265, opened 2021-11-10
Need a way to define different termination timeout for nested lifetimes


IlyaAI:
We've tried to use Lifetime's async-features in dotMemory and DPA instead of our own implementation of AsyncLifetime and faced the problem: 500ms of current default timeout isn't enough for DM/DPA activities. While DM is standalone product we can solve this by globally overriding LifetimeDefinition.WaitForExecutingInTerminationTimeoutMs. But for DPA this way looks unacceptable because DPA works side-by-side with Rider/R#. So we need a way to  define different termination timeout for nested lifetimes (associated with DPA components, for example).

korifey:
I see several ways to achieve it: 

1. The worst one - add separate int field `timeout` into LifetimeDefinition. This will make every lifetime 4 bytes bigger.
2. Use some internal bits of lifetime's state field for this timeout in logarithmic scale: 001 = 1ms, 010 = 2ms, 011 = 4ms, 100 = 8ms, etc 
3. Make cookie (thread local entity `Timeout`) instead  global static - and use it in `Terminate` method
4. Add optional parameter into Lifetime's `Terminate` method

IlyaAI:
I'm used a variation of (2): using internal bits, but instead of log scale I introduced 'semantic' timeout (aka short, long, extra-long) with actual values defined statically via LifetimeDefinition.SetTerminationTimeoutMs
See https://github.com/JetBrains/rd/pull/266

ForNeVeR:
Closed by #266.

PR #266, opened 2021-11-10
support different termination timeouts

IlyaAI:
See https://github.com/JetBrains/rd/issues/265

### DIFF
--- a/rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
+++ b/rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
@@ -24,11 +24,13 @@ namespace JetBrains.Lifetimes
 
     internal static readonly ILog Log = JetBrains.Diagnostics.Log.GetLog<Lifetime>();
     
-    [PublicAPI] internal static readonly LifetimeDefinition Eternal = new LifetimeDefinition { Id = nameof(Eternal) };
-    [PublicAPI] internal static readonly LifetimeDefinition Terminated = new LifetimeDefinition { Id = nameof(Terminated) };
+    [PublicAPI] internal static readonly LifetimeDefinition Eternal;
+    [PublicAPI] internal static readonly LifetimeDefinition Terminated;
 
     static LifetimeDefinition()
     {
+      Eternal = new LifetimeDefinition { Id = nameof(Eternal) };
+      Terminated = new LifetimeDefinition { Id = nameof(Terminated) };
       Terminated.ToCancellationToken(); //to create cts
       Terminated.Terminate();
     }
@@ -89,7 +91,32 @@ namespace JetBrains.Lifetimes
     }
     
     private const int WaitForExecutingInTerminationTimeoutMsDefault = 500;
-    [PublicAPI] public static int WaitForExecutingInTerminationTimeoutMs = WaitForExecutingInTerminationTimeoutMsDefault;  
+    [PublicAPI] public static int WaitForExecutingInTerminationTimeoutMs = WaitForExecutingInTerminationTimeoutMsDefault;
+
+    /// <summary>
+    /// Gets the actual value in milliseconds for semantic (short, long, etc) termination timeout.
+    /// </summary>
+    /// <param name="timeout">semantic timeout as defined by <see cref="LifetimeTerminationTimeout"/></param>
+    /// <returns>timeout value in milliseconds</returns>
+    [PublicAPI]
+    public static int GetTerminationTimeoutMs(LifetimeTerminationTimeout timeout) =>
+      timeout == LifetimeTerminationTimeout.Default
+        ? WaitForExecutingInTerminationTimeoutMs
+        : ourTerminationTimeoutMs[(int)timeout - 1];
+
+    /// <summary>
+    /// Sets the actual value in milliseconds for semantic (short, long, etc) termination timeout.
+    /// </summary>
+    /// <param name="timeout">semantic timeout as defined by <see cref="LifetimeTerminationTimeout"/></param>
+    /// <param name="milliseconds">timeout value in milliseconds</param>
+    [PublicAPI]
+    public static void SetTerminationTimeoutMs(LifetimeTerminationTimeout timeout, int milliseconds)
+    {
+      if (timeout == LifetimeTerminationTimeout.Default)
+        WaitForExecutingInTerminationTimeoutMs = milliseconds;
+      else
+        ourTerminationTimeoutMs[(int)timeout - 1] = milliseconds;
+    }
 
     // use real (sealed) types to allow devirtualization
     private static readonly IntBitSlice ourExecutingSlice = BitSlice.Int(20);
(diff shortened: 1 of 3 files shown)


rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
```diff
@@ -89,7 +91,32 @@ private int ThreadLocalExecuting(int increment = 0)
...
+        ? WaitForExecutingInTerminationTimeoutMs
+        : ourTerminationTimeoutMs[(int)timeout - 1];
+
+    /// <summary>
+    /// Sets the actual value in milliseconds for semantic (short, long, etc) termination timeout.
+    /// </summary>
+    /// <param name="timeout">semantic timeout as defined by <see cref="LifetimeTerminationTimeout"/></param>
+    /// <param name="milliseconds">timeout value in milliseconds</param>
+    [PublicAPI]
+    public static void SetTerminationTimeoutMs(LifetimeTerminationTimeout timeout, int milliseconds)
```
sergeyQx, line 113: Why so complicated? Why not to have 4 -values array with timeouts and always slice it?

sergeyQx (commented):
What bothers me here is that we allow anyone to create a lifetime that will be shutting down slowly and the world will have to wait.

rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
IlyaAI: I'm not sure... Do you mean why `timeout == Default ? WaitForExecutingInTerminationTimeoutMs : ourTerminationTimeoutMs[(int)timeout - 1];` instead of `ourTerminationTimeoutMs[timeout];`?
This is due to backward compatibility - the `WaitForExecutingInTerminationTimeoutMs` is a part of public API.

IlyaAI: >  ...create a lifetime that will be shutting down slowly and the world will have to wait
It is possible anyway - we can wait any time in any termination action. Our current implementation of AsyncLifetime does exactly this - waits in OnTermination action.

sergeyQx: > It is possible anyway - we can wait any time in any termination action. Our current implementation of AsyncLifetime does exactly this - waits in OnTermination action.

Lifetime waits for the timeout then issues warning and finishes its Terminate method. In your case it will wait up to 30 seconds.

IlyaAI: > why can't we use ourTerminationTimeoutMs[timeout];? when ourTerminationTimeoutMs contains 4 values?

Because the `WaitForExecutingInTerminationTimeoutMs` is a field and we can't store its actual value in array, and can't remove `WaitForExecutingInTerminationTimeoutMs` or convert to property because it is part of public API.

IlyaAI: > Lifetime waits for the timeout then issues warning and finishes its Terminate method. In your case it will wait up to 30 seconds.

If some action which was added via OnTermination() method blocks for arbitrary time, the Terminate() method will not finish till action completion. So, anyway it is possible to block termination for any arbitrary time.

sergeyQx: The change in the library is ok anyway. When we use long timout in the product and it affects product's shutdown we'll deal with it

Iliya-usov (approved):

rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
```diff
@@ -160,34 +189,57 @@ public bool AllowTerminationUnderExecution
...
-    
+
     /// <summary>
-    /// Creates toplevel lifetime definition with no parent. <see cref="Status"/> will always be <see cref="LifetimeStatus.Alive"/>.
+    /// Creates toplevel lifetime definition with no parent. <see cref="Status"/> will always be <see cref="LifetimeStatus.Alive"/>.<br/>
+    /// Created definition and all its children (if not explicitly overriden) will have specified termination timeout (see <see cref="LifetimeTerminationTimeout"/>). 
     /// </summary>
-    public LifetimeDefinition() {}    
-    
+    public LifetimeDefinition(LifetimeTerminationTimeout terminationTimeout = LifetimeTerminationTimeout.Default)
```
ForNeVeR, line 204: Please note that this breaks the API compatibility: the code compiled against older versions of Rd will likely break in runtime, and won't work without recompiling. Could you please add a parameterless override?

ForNeVeR (changes requested):

rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
```diff
@@ -160,34 +189,57 @@ public bool AllowTerminationUnderExecution
...
     
     
     #region Init
-    
+
     /// <summary>
-    /// Creates toplevel lifetime definition with no parent. <see cref="Status"/> will always be <see cref="LifetimeStatus.Alive"/>.
+    /// Creates toplevel lifetime definition with no parent. <see cref="Status"/> will always be <see cref="LifetimeStatus.Alive"/>.<br/>
+    /// Created definition and all its children (if not explicitly overriden) will have specified termination timeout (see <see cref="LifetimeTerminationTimeout"/>). 
     /// </summary>
```
korifey, line 203: Why put termination timeout as contructor parameter. It's a kind of rare thing that make API less understandable.

You can make property for LifetimeDefinition:

LifetimeTerminationTimeoutKind TerminationTimeout { get; set;}

rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
IlyaAI: ok

rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
IlyaAI: Because termination timeout inherited by nested lifetimes (if not explicitly overridden). The writable property approach much harder to implement in current design and it harder to understand in edge cases, e.g. what should happen if termination timeout set to parent lifetime while nested one already in terminating state?

korifey: When Termination Timeout is set via property, it shouldn't autimatically propagate to already existing children. Only new children can inherit Timeout.
You can imagine the most typical pattern `new LifetimeDefinition() { TerminationTimeout = Long}`

rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
korifey: It's common to have suffix "-Kind" for enums. Like  `LifetimeTerminationTimeoutKind`

IlyaAI: Ok, "-Kind" sounds good here.

rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
IlyaAI: Writable property contract doesn't suppose exactly such behavior and only way to declare it is xmldoc (but who reads the docs?) On other hand, read-only property doesn't have this behavior ambiguity. However I respect your opinion as library author and will re-work to writable property if you insist.

IlyaAI: done

rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
```diff
@@ -188,8 +188,17 @@ public bool AllowTerminationUnderExecution
...
+
+    /// <summary>
+    /// Gets or sets termination timeout kind for the definition.<br/>
+    /// The sub-definitions inherit this value at the moment of creation.
+    /// The changing of <c>TerminationTimeoutKind</c> doesn't affect already created sub-definitions.
+    /// </summary>
+    public LifetimeTerminationTimeoutKind TerminationTimeoutKind
+    {
+      get => ourTerminationTimeoutKindSlice[myState];
+      set => myState = ourTerminationTimeoutKindSlice.Updated(myState, value);
```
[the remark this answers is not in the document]
IlyaAI: Done

Iliya-usov (approved):

ForNeVeR (approved):

PR #267, opened 2021-11-13
Refactor rd-gen tests

ForNeVeR:
All the rd-gen tests have been cleaned up and migrated to the new test framework. More precise checks, more gold, more testing.


mfilippov (approved):
LGTM

PR #269, opened 2021-11-14
Serializers: dispose of a redundant comment

ForNeVeR:
Earlier, `UnsafeWriter.Cookie` was used instead of a `bookmark`, but now it isn't, so the comment is redundant: `Bookmark` is not an `IDisposable`.


ulex (approved):
👍

PR #281, opened 2021-12-28
Interpolated string handlers

Iliya-usov:


rd-net/Lifetimes/Lifetimes.csproj
```diff
@@ -15,6 +15,8 @@
         <Configurations>Debug;Release;CrossTests</Configurations>
         
         <Platforms>AnyCPU</Platforms>
+        
+        <LangVersion>10</LangVersion>
```
ForNeVeR, line 19: This shouldn't be required, since I've enabled this globally via `Directory.Build.props` in `master`.

rd-net/Test.Lifetimes/Test.Lifetimes.csproj
```diff
@@ -10,6 +10,9 @@
         <Platforms>AnyCPU</Platforms>
         <RollForward>LatestMajor</RollForward>
         <Nullable>disable</Nullable>
+        <LangVersion>10</LangVersion>
```
ForNeVeR, line 13: Same as above: this shouldn't be required.

ForNeVeR (changes requested):

ForNeVeR (approved):

PR #282, opened 2021-12-28
Fix termination during binding usov

Iliya-usov:


rd-net/RdFramework/Base/IRdBindable.cs
```diff
@@ -151,6 +151,25 @@ private static void Identify0(this IEnumerable? items, IIdentities ids, RdId id)
...
+      {
+        // can throw exception if wireable is not bound
+        return wireable.WireScheduler;
+      }
+      catch (Exception) when(!wireable.IsBound)
+      {
+        return null;
+      }
+    }
+  }
```
ForNeVeR, line 171: It isn't totally clear to me which exceptions are meant to be caught by this block. Let's discuss this.

(same in .NET counterpart)

ForNeVeR: Decided to reimplement this. At least, let's mark the "not bound"-kind exception with a separate type, and only catch this exception.

Ideally, we'd better get rid of this `catch` completely, and rearchitect this solution.

ForNeVeR (changes requested):

rd-kt/rd-framework/src/main/kotlin/com/jetbrains/rd/framework/impl/RdTask.kt
```diff
@@ -131,8 +138,13 @@ class EndpointWiredRdTask<TReq, TRes>(
...
         RdReactiveBase.logReceived.trace { "received cancellation" }
         buffer.readVoid() //nothing just a void value
-        result.setIfEmpty(RdTaskResult.Cancelled())
-        def.terminate()
+
+        val success = result.setIfEmpty(RdTaskResult.Cancelled())
+        val wireScheduler = call.wireSchedulerIfBound
+        if (success || wireScheduler == null)
+            def.terminate()
+        else if (lifetime.isAlive)
```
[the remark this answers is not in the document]
Iliya-usov: I think `executeIfAlive` is not needed here, because there is no resource we have to protect. Moreover, we should not prevent termination of lifetime during queueing, because we already want to terminate lifetime.
`if (lifetime.isAlive)` is just an optimization to avoid useless queueing if lifetime is already terminated

ForNeVeR (approved):

PR #286, opened 2022-01-13
[CallerExpressionArgument] for .NotNull()

controlflow:
It is really nice to have `[CallerExpressionArgument]` for `.NotNull()` extension method too, but to do so we need replace the current set of `NotNull` overloads:
```c#
T NotNull<T>(this T? value) where T : class;
T NotNull<T>(this T? value) where T : struct;
T NotNull<T>(this T?, string message) where T : class;
T NotNull<T>(this T?, string message) where T : struct;
```
With the reduced set of method with optional parameters:
```c#
T NotNull<T>(this T?, [CallerExpressionArgument] string? message = null) where T : class;
T NotNull<T>(this T?, [CallerExpressionArgument] string? message = null) where T : struct;
```

Unfortunately, this will break the binary compatibility.
Probably the transition between `2021.x` and `2022.x` is the best time to do this kind of change.

Sorry for dirty looking diff, I've reordered a few members to look uniform with `Assert`/`AssertNotNull` overloads.

### DIFF
--- a/rd-net/Lifetimes/Diagnostics/Assertion.cs
+++ b/rd-net/Lifetimes/Diagnostics/Assertion.cs
@@ -154,7 +154,7 @@ namespace JetBrains.Diagnostics
     {
       if (condition == null)
       {
-        Fail(message ?? "");
+        Fail(message ?? "Not null expected");
       }
     }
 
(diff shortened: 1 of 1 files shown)


hypersw:
I second this, but I'm not sure how breaking this change might be.
NOTE that any existing code should recompile smoothly, just binary compat would break.
And binary compat does not usually hold for our stuff between major releases, including RD protocol.

PR #303, opened 2022-05-11
Add ExecutionWasNotCancelledByTimeout signal

Iliya-usov:
Add ExecutionWasNotCancelledByTimeout signal to improve diagnostics (dump thread) dumps in tests

### DIFF
--- a/rd-net/Test.Lifetimes/Lifetimes/LifetimeTest.cs
+++ b/rd-net/Test.Lifetimes/Lifetimes/LifetimeTest.cs
@@ -214,6 +214,7 @@ namespace Test.Lifetimes.Lifetimes
       const string expectedWarningText = "can't wait for `ExecuteIfAlive` completed on other thread";
       const string expectedExceptionText = "ExecuteIfAlive after termination of";
       bool warningReceived = false, exceptionReceived = false;
+      var executionWasNotCancelledByTimeoutReceived = false;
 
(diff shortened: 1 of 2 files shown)


rd-net/Test.Lifetimes/Lifetimes/LifetimeTest.cs
```diff
@@ -228,7 +229,20 @@ void LoggerHandler(LeveledMessage message)
           () => TestLogger.ExceptionLogger.Handlers -= LoggerHandler
           );
 
+
         var lifetimeDefinition = lifetime.CreateNested();
+
+        var def2 = lifetime.CreateNested();
```
ForNeVeR, line 235: Please do `using var`. Otherwise, this test's failure may affect further ones.

ForNeVeR (changes requested):

rd-net/Test.Lifetimes/Lifetimes/LifetimeTest.cs
Iliya-usov: it is not necessary because parent `lifetime` will be terminated when the test is finished

ForNeVeR: Ah alright

ForNeVeR (approved):

PR #307, opened 2022-05-25
Extract RdContext into a base class to make value propert virtual

nerzhulart:


rd-kt/rd-gen/src/main/kotlin/com/jetbrains/rd/generator/nova/Types.kt
```diff
@@ -447,7 +450,7 @@ abstract class Toplevel(pointcut: BindableDeclaration?) : BindableDeclaration(po
     fun flags(body: Enum.() -> Unit) = enum( body).apply { flags = true }
 
     fun internScope(name: String = "") = InternScope(this, name)
-    fun context(type: INonNullableScalar) = append(Context.Generated(this, "", type)) {}
```
ForNeVeR, line 450: Could we please have an old function (which will delegate to `threadLocalContext`) as `@Deprecated`? Breaking changes are always very painful, and it's hard to apply them across all the codebase.

ForNeVeR (changes requested):
In principle, LGTM. But let's wait for a review from @Iliya-usov.

ForNeVeR (approved):

Iliya-usov (approved):

PR #312, opened 2022-06-08
[Kotlin] Replace RdContext.value_set with updating cookie pattern. It…

nerzhulart:
… needed for more proper lifetime control of values (ClientId migration to Idea platform contexts)


nerzhulart:
I can restore setter and mark it deprecated for backward compat, but I guess we also can remove it because the fix in Idea/Rider sources is simple

rd-kt/rd-framework/src/main/kotlin/com/jetbrains/rd/framework/impl/ProtocolContexts.kt
```diff
@@ -137,23 +137,20 @@ class ProtocolContexts(val serializationCtx: SerializationCtx) : RdReactiveBase(
     fun readMessageContextAndInvoke(buffer: AbstractBuffer, action: () -> Unit) {
         val numContextValues = buffer.readShort().toInt()
         assert(counterpartHandlers.size >= numContextValues) { "We know of ${counterpartHandlers.size} remote keys, received $numContextValues instead" }
-        val oldValues = ArrayList<Any?>(numContextValues)
+        val valueRestorers = mutableListOf<AutoCloseable>()
```
Iliya-usov, line 140: I thinlk it would be better to use `List(numContextValues) { i -> ... })`

PR #313, opened 2022-06-08
[.NET] Get rid of RdContext.Value setter and PushValue/PopValue and r…

nerzhulart:
…eplace them with special cookie which restores the value on disposing.

### DIFF
--- a/rd-net/RdFramework/Base/ISingleContextHandler.cs
+++ b/rd-net/RdFramework/Base/ISingleContextHandler.cs
@@ -1,12 +1,12 @@
+using System;
 using JetBrains.Serialization;
 
 namespace JetBrains.Rd.Base
 {
   internal interface ISingleContextHandler
   {
-    void ReadValueAndPush(SerializationCtx context, UnsafeReader reader);
-    void PopValue();
-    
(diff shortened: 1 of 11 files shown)


nerzhulart:
I can restore the setter and mark it Obsole for backward compat, but I guess we can remove it because it isn't used in Rider/R# codebase

rd-net/RdFramework/Base/ISingleContextHandler.cs
```diff
@@ -1,12 +1,12 @@
+using System;
 using JetBrains.Serialization;
 
 namespace JetBrains.Rd.Base
 {
   internal interface ISingleContextHandler
   {
-    void ReadValueAndPush(SerializationCtx context, UnsafeReader reader);
```
ForNeVeR, line 7: _(Minor)_ We could win a couple of boxing/indirect calls here by exposing `UpdateValueCookie` through this interface and making it a `struct`, but I'm not sure it's worth it.

ForNeVeR (approved):

rd-net/RdFramework/Base/ISingleContextHandler.cs
ForNeVeR: Decided to let it be for now. I'll think more about improvements in this area later.

PR #323, opened 2022-07-05
Temporary workaround for RIDER-77529

real-mikhail:
Change allows to increase buffer size for messages in rd protocol.


mfilippov:
@ForNeVeR We need it in 222

ForNeVeR:
Ok, will release for 222 ASAP.

PR #324, opened 2022-07-08
Better workaround for RIDER-77529

real-mikhail:
Allows to set the maximum message length from the caller code


mfilippov (approved):
LGTM

PR #325, opened 2022-07-11
Improvments

Iliya-usov:


ForNeVeR:
`System.Threading.Channels` NuGet package depends on `System.Threading.Tasks.Extensions`, which depends on `System.Runtime.CompilerServices.Unsafe`, which I remember as being problematic.
image

rd-net/Lifetimes/Lifetimes.csproj
```diff
@@ -18,6 +18,10 @@
         
     </PropertyGroup>
     
+    <PropertyGroup Condition="'$(TargetFramework)' == 'net35'">
+        <WarningsNotAsErrors>CS8632</WarningsNotAsErrors>
+    </PropertyGroup>
+
```
ForNeVeR, line 24: Should this actually go to the `Directory.Build.props`?

Actually wait a minute, our `Directory.Build.props` already has `<NoWarn>nullable</NoWarn>` for this TFM. Doesn't it work?

ForNeVeR (changes requested):

ForNeVeR:
Turns out we have already included all of the `System.Threading.Channels`' dependencies, so I retract my complaint. Using the package is okay.

rd-net/Lifetimes/Lifetimes.csproj
Iliya-usov: Probably it doesn't work for Rider.  Rider shows a lot of errors in SWEA.

ForNeVeR: Let's unvestigate; I was unable to reproduce.

ForNeVeR (commented):
Sorry, but no. RiderFlow gets broken with this new dependency. We cannot allow us to include `System.Threading.Channels`.

rd-net/Lifetimes/Lifetimes.csproj
```diff
@@ -59,4 +59,8 @@
         <None Include="PackageIcon.png" Pack="true" PackagePath="" />
     </ItemGroup>
 
+    <ItemGroup Condition="'$(TargetFramework)' != 'net35'">
+      <PackageReference Include="System.Threading.Channels" Version="6.0.0" />
+    </ItemGroup>
```
ForNeVeR, line 64: We don't need this anymore, right?

ForNeVeR (approved):

rd-net/Lifetimes/Lifetimes.csproj
Iliya-usov: yep

PR #326, opened 2022-07-15
Reduce memory traffic

Iliya-usov:


ForNeVeR (changes requested):

rd-net/Lifetimes/Util/SingleThreadObjectPool.cs
```diff
@@ -0,0 +1,71 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using JetBrains.Diagnostics;
+
+namespace JetBrains.Util;
+
+public class SingleThreadObjectPool<T>
```
ulex, line 8: What is the idea of limiting this class to only one thread? Why not just use the ConcurrentBox as the backend?
It would be nice to have more comments on this class. It seems to be useful for general purpose

ForNeVeR: Originally, it was called `SimpleObjectPool`, so the idea was to make it simple 😅 

Since we only use it for the thread-static context now, I think this is justified.

ForNeVeR (approved):

rd-net/Lifetimes/Util/SingleThreadObjectPool.cs
Iliya-usov: Yep, we want to have a pool without the overhead of synchronization and heap allocation.

ulex: @Iliya-usov in most cases there is no synchronization overhead in the ConcurrentBox. And I don't understand the argument regarding heap overhead because the proposed solution uses Stack<T>

@ForNeVeR I'm not suggested make it "complex". My idea was to make this class reusable by replacing one collection with another and adding a comment about this class.

ForNeVeR: Ok, I have nothing against generalizing this over any kind of collection.

Iliya-usov: By heap allocation overhead, I meant the fact that concurrent collections in the .net framework often allocate new objects.. For example, `ConcurrentBag` creates a new object for every call to the `Add` method.

PR #327, opened 2022-07-16
RIDER-77529: threading improvement

ForNeVeR:
I believe the Council has decided that the `@Volatile` is really required there.


ForNeVeR:
cc @real-mikhail, that's what we've been discussing.

DedSec256 (approved):

Iliya-usov (approved):

PR #329, opened 2022-07-19
Fix races in LifetimeDefinition

ulex:
This pull request fixes two issues regarding CancellationTokenSource in LifetimeDefinition:

  * First issue with destruction may lead to the null reference exception when cancellation token is being requested on terminating lifetime (kudos to @Denis-Korneev for noticing the failed test on teamcity and suggested fix)
 * Second issue is that not all subscribers are guaranteed to be notified about cancellation because of the error in the initialization of myCts.


ForNeVeR (approved):
LGTM, but waiting for thorough review from @Iliya-usov.

Iliya-usov (approved):

PR #330, opened 2022-08-04
Fix NextValueAsync

Iliya-usov:


rd-net/Lifetimes/Collections/Viewable/ReactiveEx.cs
```diff
@@ -413,12 +413,16 @@ public static Task<T> NextValueAsync<T>(this ISource<T> source, Lifetime lifetim
 
     public static Task<T> NextValueAsync<T>(this ISource<T> source, Lifetime lifetime, Func<T, bool> condition)
     {
-      var tcs = lifetime.CreateTaskCompletionSource<T>();
-      source.AdviseOnce(lifetime, v =>
+      var tcs = new TaskCompletionSource<T>();
+      var definition = lifetime.CreateNested();
+      definition.SynchronizeWith(tcs);
+
+      source.Advise(definition.Lifetime, v =>
```
van800, line 420: Why not using AdviseUntil, instead of definition.SynchronizeWith(tcs)?

Iliya-usov: to create only 1 nested lifetime

ForNeVeR (changes requested):
Could you please explain the issue this PR is solving?

Preferably in the form of a unit test 😏

ForNeVeR (approved):
Thanks!

PR #343, opened 2022-09-12
Improvments

Iliya-usov:
Introduce lifetimedCoroutineScope
Deprecate launchChild and startChildAsync api (It is now preferable to use only launch and async (without lifetime) or use lifetimedCoroutineScope
Add overloads for LogEx with String Interpolated Handlers

### DIFF
--- a/rd-kt/rd-framework/src/main/kotlin/com/jetbrains/rd/framework/util/LifetimeCoroutineUtil.kt
+++ b/rd-kt/rd-framework/src/main/kotlin/com/jetbrains/rd/framework/util/LifetimeCoroutineUtil.kt
@@ -111,6 +111,10 @@ fun Lifetime.createTerminatedAfter(duration: Duration, terminationContext: Corou
         }
     }
 
+
+/**
+ * Creates a [coroutineScope] that will be cancelled on the passed lifetime termination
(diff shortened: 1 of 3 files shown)


rd-kt/rd-framework/src/main/kotlin/com/jetbrains/rd/framework/util/LifetimeCoroutineUtil.kt
```diff
@@ -108,4 +109,11 @@ fun Lifetime.createTerminatedAfter(duration: Duration, terminationContext: Corou
             delay(duration.toMillis())
             nested.terminate()
         }
-    }
\ No newline at end of file
+    }
+
+suspend fun <T> lifetimedCoroutineScope(lifetime: Lifetime, action: suspend CoroutineScope.() -> T) = coroutineScope {
```
ForNeVeR, line 114: Some documentation wouldn't hurt here IMO. Something like "creates a coroutine scope that will be cancelled on lifetime termination, or parent scope cancellation".

ForNeVeR (changes requested):

ForNeVeR (approved):

PR #354, opened 2022-10-24
Improge Rd Logger

Iliya-usov:
Fix stofl in getLogger
Reduce memory consumption
Do not create listener for all SwitchLogger (use cached LoggerMediator) Fix SwitchLogger mem leak
Fix race condition when getLogger is called at the same moment as Statics<ILoggerFactory>().push

### DIFF
--- a/rd-kt/rd-core/src/main/kotlin/com/jetbrains/rd/util/Logger.kt
+++ b/rd-kt/rd-core/src/main/kotlin/com/jetbrains/rd/util/Logger.kt
@@ -1,8 +1,6 @@
 package com.jetbrains.rd.util
 
 import com.jetbrains.rd.util.lifetime.Lifetime
-import com.jetbrains.rd.util.lifetime.onTermination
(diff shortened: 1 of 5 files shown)


rd-kt/rd-core/src/main/kotlin/com/jetbrains/rd/util/Logger.kt
```diff
@@ -32,25 +30,35 @@ interface ILoggerFactory {
...
 
+            pair = Pair(currentFactory, currentFactory.getLogger(category))
+            cachedPair = pair
+        }
+    }
+
+    private class Pair(val factory: ILoggerFactory, val logger: Logger)
 }
 
+internal object LoggerFactoryStatic {
```
ForNeVeR, line 58: _(Minor)_
```suggestion
private object LoggerFactoryStatic {
```

ForNeVeR (approved):

PR #358, opened 2022-11-06
rd-net: Improve `RdTask` factory methods usage experience

seclerp:
Currently, to return the successful result you should provide a full specific type when dealing with `RdTask<T>.Successful`. It could be quite massive:
```csharp
return RdTask<IEnumerable<SomeResultModel>>.Successful(results);
```

This PR introduces a special non-generic `RdTask` class with generic types moved to method signatures. This allows to use the power of the C# compiler to infer the type from a method argument and simplifies usage:
```csharp
return RdTask.Successful(results);
```

`Faulted` and `Cancelled` are also introduced in `RdTask` to remove possible user confusion.

### DIFF
--- a/rd-net/RdFramework.Reflection/ProxyGenerator.cs
+++ b/rd-net/RdFramework.Reflection/ProxyGenerator.cs
@@ -214,7 +214,8 @@ namespace JetBrains.Rd.Reflection
       if (IsSync(method))
       {
         // Create RdTask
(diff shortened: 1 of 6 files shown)


ForNeVeR (approved):

ForNeVeR:
@ulex, please take a look at the changes in `ProxyGenerator`.

ulex:
Sorry for the delay, was on vacation. LGTM

PR #359, opened 2022-11-09
Rd gen duration support

estrizhok:
Added support for TimeSpan to rd_gen to be used in models. Represented by kotlin.time.Duration in kotlin and custom TimeSpan in cpp. I haven't been able to use std::chrono::duration in cpp, because it has complicated conversion rules between different time unit: https://www.reddit.com/r/cpp/comments/7otsjh/comment/dscd79y/


ForNeVeR (approved):
👍

PR #360, opened 2022-11-10
Fix unaligned memory access in UnsafeReader.ReadDouble() and UnsafeWriter.Write(double) on Linux ARM32

k15tfu:
Fix unaligned memory access in UnsafeReader.ReadDouble() and UnsafeWriter.Write(double) on Linux ARM32

https://youtrack.jetbrains.com/issue/RSRP-490138


ForNeVeR (approved):

ulex:
What about all other write/read methods? they could all read from unaligned addresses?

PR #361, opened 2022-11-11
Upgrade target framework net461 to net472

k15tfu:


ForNeVeR (changes requested):
Let's skip 471 and upgrade to net472?

ForNeVeR (approved):

PR #365, opened 2022-12-12
RdReflection proxy: support for more than 7 arguments in method calls

ulex:


rd-net/Test.RdFramework.Generator/ArgsProxyTests.cs
```diff
@@ -0,0 +1,57 @@
...
+      string source = $@"// <auto-generated/>
+using JetBrains.Diagnostics;
+using JetBrains.Rd.Reflection;
+using Microsoft.CodeAnalysis;
+using NUnit.Framework;
+namespace Test.RdFramework.Reflection;
+
+[TestFixture]
+public partial class ProxyGeneratorCalls
+{{
```
ForNeVeR, line 24: Note: you can use C# 11 `$$"""` feature to avoid these ugly `{{`, I think 😄

ForNeVeR (approved):

rd-net/Test.RdFramework.Generator/ArgsProxyTests.cs
ulex: I guess this requires update to 7.0

ForNeVeR: Only to SDK 7, but you're right. I'll note that in my backlog.

PR #380, opened 2023-02-01
Assertion review

ulex:
The most of the assertions were disabled in release build previously
    using JET_MODE_ASSERT define. The code was written with the assumption
    that Assertion.Assert methods invocations are removed in release mode.

    This is no longer truth. The NuGet packages builds with the
    JET_MODE_ASSERT enabled. To reduce performance impact of the Assertion
    I added additional checks for Mode.Assertion to help JIT to remove
    the branch with the check (in perf-critical code)

    Additionally, in non-performance critical places, I converted
    Assertions.Assert to Assertion.Require method, which is not marked by
    ConditionalCompilation.

### DIFF
--- a/rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
+++ b/rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
@@ -503,8 +503,8 @@ namespace JetBrains.Lifetimes
       if (Mode.IsAssertion) Assertion.Assert(ourMutexSlice[myState] == false, "{0}: mutex must be released in this point", this);
       //no one can take mutex after this point
 
-      var resources = myResources;
-      if (Mode.IsAssertion) Assertion.Assert(resources != null, "{0}: `resources` can't be null on destructuring stage", this);
+      var resources = myResources!;
+      if (Mode.IsAssertion) Assertion.AssertNotNull(resources, "{0}: `resources` can't be null on destructuring stage", this);
       
       for (var i = myResCount - 1; i >= 0; i--)
       {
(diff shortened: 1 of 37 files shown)


ulex:
@ForNeVeR sorry for stuffing two commits in one review, second one is minor change in RdReflection

rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
```diff
@@ -499,12 +499,12 @@ private void MarkCancelingRecursively()
...
       var status = Status;
-      Assertion.Assert(status == LifetimeStatus.Terminating, "{0}: bad status for destructuring start", this);
-      Assertion.Assert(ourMutexSlice[myState] == false, "{0}: mutex must be released in this point", this);
+      if (Mode.IsAssertion) Assertion.Assert(status == LifetimeStatus.Terminating, "{0}: bad status for destructuring start", this);
+      if (Mode.IsAssertion) Assertion.Assert(ourMutexSlice[myState] == false, "{0}: mutex must be released in this point", this);
       //no one can take mutex after this point
 
       var resources = myResources;
-      Assertion.Assert(resources != null, "{0}: `resources` can't be null on destructuring stage", this);
+      if (Mode.IsAssertion) Assertion.Assert(resources != null, "{0}: `resources` can't be null on destructuring stage", this);
```
ForNeVeR, line 507: _(Minor)_ Let's add `!` below, in the `try-catch` block, to shut up the inspection?

ForNeVeR (approved):

rd-net/Lifetimes/Lifetimes/LifetimeDefinition.cs
ulex: Thanks, fixed!

PR #384, opened 2023-02-10
Add infersharp analisys

mfilippov:


ForNeVeR (approved):

ForNeVeR:
From the items it found, we can conclude InferSharp is not that useful on our code.

image

These two are just wrong: the resources get added to the log lifetime.

image

At least this one is also incorrect, all the reads are done from under a lock (sometimes indirectly).

I thought their cross-procedural analysis is better.

PR #398, opened 2023-04-23
Build: remove JaCoCo

ForNeVeR:
**Why?**

JaCoCo takes time to execute and resources to maintain, but we don't use its results in any way.


mfilippov (approved):
LGTM

PR #402, opened 2023-05-05
UnsafeWriter/UnsafeReader rework

controlflow:
* UnsafeWriter: introduce `WriteInt32()`-like APIs instead of one highly-overloaded single `Write()` method to avoid bugs annoying in serializers when you accidentally choose wrong overload (for example, expect to serialize some byte-sized value, while `Write(0)` invocation actually writes 4 bytes through `Int32` overload). Those APIs also mirror `UnsafeReader` API surface. I've introduced [CodeTemplate] migrators to simplify the migrations. Binary compatibility is preserved.
* UnsafeWriter: add new API to have the ability to allocate some amount of memory and write less than this amount through raw pointer - useful to allow string serialization with custom encoding (see new test).
* UnsafeWriter/UnsafeReader: two new APIs to read/write compressed Int32 values
* Get rid of unnecessary `AggressiveInlining` usage
* Fix warnings and cleanup the code

### DIFF
--- a/rd-net/Lifetimes/Annotations/CodeAnnotations.cs
+++ b/rd-net/Lifetimes/Annotations/CodeAnnotations.cs
@@ -564,4 +564,202 @@ namespace JetBrains.Annotations
     /// </summary>
     [AttributeUsage(AttributeTargets.Parameter)]
     internal sealed class RegexPatternAttribute : Attribute { }
+
+    /// <summary>
+    /// <para>
+    /// Defines the code search template using the Structural Search and Replace syntax.
+    /// It allows you to find and, if necessary, replace blocks of code that match a specific pattern.
+    /// Search and replace patterns consist of a textual part and placeholders.
+    /// Textural part must contain only identifiers allowed in the target language and will be matched exactly (white spaces, tabulation characters, and line breaks are ignored).
+    /// Placeholders allow matching variable parts of the target code blocks.
+    /// A placeholder has the following format: $placeholder_name$- where placeholder_name is an arbitrary identifier.
+    /// </para>
+    /// <para>
+    /// Available placeholders:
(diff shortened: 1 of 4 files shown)


controlflow:
> Could you please make sure the JIT can handle double inlining well?

https://sharplab.io/#v2:EYLgHgbALANALiAhgZwLYB8ACAmAjAWAChMAGAAk1wDoAlAVwDs4BLVAUyoGEB7VAB2YAbNgCcAyqIBuzAMZtkAbiJEcFXAHYyAbyJk1EClDIBZRMwYAKAJTbdetQE4LDNgHcyAMW7drVAOoizHBsuBa4VlZKhPaOzm5kAEKIIr4BQSFhEVF6AL5EeYQq2J7ettFkANrGbHAAFtwAJgCS/IIW1XWNLXyCAPJ8LNwMyFQAggDm4yLyyMySbE0Mgubm41YAunaYAMxkgkPjZGnBofsMh2A2OuV6mJrHbNgWl9lkBTF2VTX1za3t311Wv1BsMxpNpshZvNFssGKsNnYzocHk8kWRLmUYnd0WQANRkEhgAAiAFFRkSEiSSR5XgUCkVEslMRRdmiHqcDuirnZbvdAsEni87O97IjOSiLGiMdcsZowHiCcSyRSqTThfkgA

> Most of existing code uses polymorphic Write method, and it wouldn't be great if the call chain `Write(byte)->WriteByte()->actual write` isn't inlined completely.

It's not "polymorphic", it is overloaded.

> There is exiting `UnsafeMarshallersTest.Bdn` to test for about the same. Except that, LGTM

ulex:
Thanks for checking!

ulex (approved):

PR #442, opened 2023-10-19
Fix memory leak

Iliya-usov:


rd-kt/rd-core/src/main/kotlin/com/jetbrains/rd/util/lifetime/RLifetime.kt
```diff
@@ -642,7 +642,20 @@ operator fun Lifetime.plusAssign(action : () -> Unit) = onTermination(action)
...
-fun Lifetime.intersect(lifetime: Lifetime): LifetimeDefinition = Lifetime.defineIntersection(this, lifetime)
+fun Lifetime.intersect(lifetime: Lifetime): Lifetime {
+    if (lifetime === this)
+        return this
+
+    if (lifetime.isEternal)
+        return lifetime
+
+    if (this.isEternal)
+        return this
```
ForNeVeR, line 653: If I'm not mistaken, this should be the other way around.

ForNeVeR (changes requested):

ForNeVeR (approved):

PR #459, opened 2024-01-02
Socket processing improvements.

mirasrael:
1. Add SimpleSocketSender friend class for SimpleSocket for avoiding concurrent state modifications during racing receive and send operations.
2. Don't log warning if socket shutdown initiated by client.
3. Make StatTimer thread local.


rd-cpp/thirdparty/clsocket/src/PassiveSocket.cpp
```diff
@@ -213,10 +209,9 @@ CActiveSocket *CPassiveSocket::Accept() {
     // Wait for incoming connection.
     //--------------------------------------------------------------------------
     if (pClientSocket != NULL) {
-        CSocketError socketErrno = SocketSuccess;
+        CSocketError socketErrno;
```
ForNeVeR, line 212: Are you sure we don't want to initialize this variable?

rd-cpp/thirdparty/clsocket/src/SimpleSocketSender.h
```diff
@@ -0,0 +1,40 @@
...
+    std::shared_ptr<CSimpleSocket> m_socket;
+    CSimpleSocket::CSocketError m_error;
+
+public:
+    explicit CSimpleSocketSender(const std::shared_ptr<CSimpleSocket>& socket) : m_error(CSimpleSocket::SocketSuccess)
+    {
+        if (socket->m_nSocketType == CSimpleSocket::CSocketType::SocketTypeTcp)
+            m_socket = socket;
+        else
+            m_socket = std::make_shared<CSimpleSocket>();
```
ForNeVeR, line 19: Not sure I understand this verse. Why does it create a new socket if a passed socket is not a TCP one?

rd-cpp/thirdparty/clsocket/src/StatTimer.h
```diff
@@ -62,6 +62,8 @@
 
 #include "Host.h"
 
+#include <algorithm>
```
ForNeVeR, line 65: _(Minor)_ Why do we need `<algorithm>` here?

ForNeVeR (changes requested):

rd-cpp/thirdparty/clsocket/src/StatTimer.h
mirasrael: I believe I left it after one of refactoring passes. Can be removed.

rd-cpp/thirdparty/clsocket/src/SimpleSocketSender.h
mirasrael: Because only TCP socket is supported. Technically new socket is always in disconnected state and every attempt to send to this socket will fail because socket isn't valid. We can throw exception instead in that case as an alternative.

rd-cpp/thirdparty/clsocket/src/PassiveSocket.cpp
mirasrael: It is guaranteed to be initialised below either 225 or 239 line. So assignment here is noop.

rd-cpp/thirdparty/clsocket/src/SimpleSocketSender.h
ForNeVeR: I would suggest throwing then.

ForNeVeR (approved):

PR #465, opened 2024-01-11
Additional diagnostis for lifetimes

Iliya-usov:


ForNeVeR (approved):
LGTM, plese investigate what's that with tests though.

ForNeVeR (approved):

PR #475, opened 2024-03-14
Improve 'toString' representation of protocol entities

JB-Dmitry:
Updates single-line PrettyPrinter logic to insert a space between joined lines.

'toString' samples
Before the fix: `LxMouseEvent (type = EnteredxAbs = 640yAbs = 512clickCount = 0causedByTouchEvent = falsebutton = 0popupTrigger = falsemodifiers = 0)`
After the fix: `LxMouseEvent ( type = Moved xAbs = 870 yAbs = 632 clickCount = 0 causedByTouchEvent = false button = 0 popupTrigger = false modifiers = 0 )`


ForNeVeR (approved):
Thanks!

PR #480, opened 2024-04-19
Improve messages in Logger.catch

Iliya-usov:


rd-kt/rd-core/src/main/kotlin/com/jetbrains/rd/util/Logger.kt
```diff
@@ -109,12 +109,16 @@ inline fun Logger.error(msg: () -> Any?) = log(LogLevel.Error, msg)
...
+}
+
 inline fun Logger.catch(comment: String?, action:() -> Unit) {
     try {
         action()
     } catch (e : Throwable) {
-        val sfx = if (comment.isNullOrBlank()) "" else ": $comment"
-        error("Catch$sfx", e)
+        val sfx = if (comment.isNullOrBlank()) e.message else ": $comment `Exception message: ${e.message}`"
+        error("Catch $sfx", e)
```
ForNeVeR, line 121: _(Minor)_ Let's make this somewhat uniform.
```suggestion
        val sfx = "${e.javaClass.name} ${e.message}" + if (comment.isNullOrBlank()) "" else " ($comment)"
        error("Catch $sfx", e)
```
This way, the messages will be one of:
- `Catch java.lang.RuntimeException Error message goes here`
- `Catch java.lang.NullPointerException` (note that this one rarely has a meaningful message)
- `Catch java.lang.RuntimeException Error message goes here (user comment goes here)`

What do you think?

ForNeVeR (approved):

PR #481, opened 2024-05-16
Rd-gen docs

ForNeVeR:
Enough is enough.

I propose to merge this as-is (after verifying it's correct) and fill up the remaining parts later.

### DIFF
--- a/docs/rd-gen.md
+++ b/docs/rd-gen.md
@@ -168,7 +168,38 @@ field("nls_nullable_field", PredefinedType.string.nullable.attrs(KnownAttrs.Nls)
 ```
 
 #### Interning
-TODO
+It's possible to reduce protocol traffic by enabling interning of certain objects. In the model definition, you can call `.interned(scope)` on any scalar definition, and that will start interning the values. The interned values are only sent the first time they appear in a scope, and in future appearances, only their id is sent.
+
+`scope` is the scope in that you want the values to be interned. For globally-interned values, use global `ProtocolInternScope`. Additionally, you can declare an intern scope in any top-level definition.
+
+A user-declared scope should be rooted in some model. TODO: Figure out the details
+
+Some examples:
+
+```kotlin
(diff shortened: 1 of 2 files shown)


docs/rd-gen.md
```diff
@@ -0,0 +1,249 @@
...
+   val DerivedClass = classdef extends BaseClass {
+       field("derivedField", PredefinedType.bool)
+   }
+   ```
+
+   Classes may be open for external inheritance as well. For that, use `openclass` instead of `classdef`.
+3. `interfacedef`: declares an interface that may be implemented by other types. Example:
+   ```kotlin
+   val Interface = interfacedef {
+       call("foo", PredefinedType.void, PredefinedType.void)
```
Iliya-usov, line 114: I'm not sure it works. Because we don't support fields for interfaces. How is `call` supposed to work with structures? I think it's a bug that we allow calls here. 
As far as I know, we only support `method` here, but it's dark magic that we don't use it.

docs/rd-gen.md
```diff
@@ -0,0 +1,249 @@
...
+}
+```
+
+### Members
+Model types may contain a variety of members:
+- constants: `const("name", type`
+- read-only fields: `field("name", type, value)`
+- signals (bi-directional): `signal("name", signalValueType)`
+- sources (the signalling side of a signal): `source("name", signalValueType)`
+- sinks (the receiving side of a signal): `sink("name", signalValueType)`
```
ForNeVeR, line 211: Is this correct? Check.

docs/rd-gen.md
ForNeVeR: Not sure if this is supported actually. The intended use is indeed via `method`. We only have tests for `method`. I'll fix this section.

docs/rd-gen.md
ForNeVeR: Seems it is , but it's kinda confusing.

PR #488, opened 2024-07-29
Add Contains and IndexOf methods for consistency with System.Collections.Generic.List<>

DenisPimenov:
Hi! I found that _CompactList_ does not contain _Contains_ and _IndexOf_ methods, which makes it impossible to use in situations where it replaces _System.Collections.Generic.List<>_. I implemented them in a similar way to the existing _LastIndexOf_. I would be glad if you can accept this changes or tell me what needs to be improved


ForNeVeR (approved):
Thanks!

ulex (approved):

PR #490, opened 2024-08-07
Do not log CancellationException in Logger.catch()

nerzhulart:


ForNeVeR (approved):
Yep, looks okay. Waiting for a second opinion from @Iliya-usov.

PR #499, opened 2024-10-22
Unix sockets support

ArtManyak:


rd-net/RdFramework.Reflection/RdFramework.Reflection.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFrameworks>netstandard2.0;net35;net472</TargetFrameworks>
+        <TargetFramework>net8.0</TargetFramework>
```
ForNeVeR, line 4: Before merging, we'll have to restore the compatibility with the older frameworks.

rd-net/RdFramework.Reflection/ReflectionRdActivator.cs
```diff
@@ -148,7 +148,8 @@ private object ActivateRd(Type type)
       object instance;
       try
       {
-        instance = Activator.CreateInstance(implementingType);
+        instance = Activator.CreateInstance(implementingType) 
+                   ?? throw new InvalidOperationException($"Unable to create instance of: {implementingType.ToString(true)}");
```
ForNeVeR, line 152: _(MInor)_ Sorry, why this change? It is not practical to expect `null` from `Activator.CreateInstance`: it will only return `null` if you pass a `Nullable` as the `implementingType`. It will throw an exception by itself otherwise; it doesn't need our additional assertion here.

rd-net/RdFramework.Reflection/ReflectionSerializerVerifier.cs
```diff
@@ -148,7 +148,10 @@ bool IsValidArray()
         if (!typeInfo.IsArray) return false;
         if (typeInfo.GetArrayRank() != 1) return false;
 
-        var arrayType = typeInfo.GetElementType().GetTypeInfo();
+        var elementType = typeInfo.GetElementType();
+        if (elementType == null) return false;
```
ForNeVeR, line 152: Should never be `null` here, I'd prefer this code to fail in such a case (as it was before) instead of trying to handle the logic error.

rd-net/RdFramework.Reflection/SerializerReflectionUtil.cs
```diff
@@ -68,13 +68,13 @@ internal static FieldInfo[] GetBindableFields(TypeInfo typeInfo)
       return list.ToArray();
     }
 
-    private static IEnumerable<FieldInfo> GetFields(Type type, Type baseType)
+    private static IEnumerable<FieldInfo> GetFields(Type? type, Type baseType)
```
ForNeVeR, line 71: Huh, should `type` really be nullable here? Who calls this on a null type?

rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -522,9 +523,12 @@ protected virtual bool AcceptHandshake(Socket socket)
     public class Client : Base
     {
       public Client(Lifetime lifetime, IScheduler scheduler, int port, string? optId = null) :
-        this(lifetime, scheduler, new IPEndPoint(IPAddress.Loopback, port), optId) {}
+        this(lifetime, scheduler, EndPointWrapper.CreateIpEndPoint(IPAddress.Loopback, port), optId) {}
+      
+      public Client(Lifetime lifetime, IScheduler scheduler, string path, string? optId = null) :
+        this(lifetime, scheduler, EndPointWrapper.CreateUnixEndPoint(path), optId) {}
```
ForNeVeR, line 529: _(Minor)_ Since this an unusual use case, I'd prefer to have something other than a `string` in here. Let's maybe make a separate struct, like this?
```csharp
public struct UnixSocketConnectionParams
{
  public string Path { get; set; }
}
```
and then
```
public Client(Lifetime lifetime, IScheduler scheduler, UnixSocketConnectionParams connectionParams, string? optId = null)
```

WDYT?

rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -639,25 +643,32 @@ public Server(Lifetime lifetime, IScheduler scheduler, Socket serverSocket, stri
       private Server(Lifetime lifetime, IScheduler scheduler, string? optId = null) : base("ServerSocket-"+(optId ?? "<noname>"), lifetime, scheduler)
       {}
 
-      public static Socket CreateServerSocket(IPEndPoint? endPoint)
+      public static Socket CreateServerSocket(EndPointWrapper? endPointWrapper)
       {
-        Protocol.InitLogger.Verbose("Creating server socket on endpoint: {0}", endPoint);
+        Protocol.InitLogger.Verbose("Creating server socket on endpoint: {0}", endPointWrapper?.EndPointImpl);
+        // by default we will use IPEndpoint?
```
ForNeVeR, line 649: _Answer:_ yeah, let's allow that behavior to continue.

rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -639,25 +643,32 @@ public Server(Lifetime lifetime, IScheduler scheduler, Socket serverSocket, stri
...
-        var serverSocket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
+        var serverSocket = new Socket(endPointWrapper.AddressFamily, endPointWrapper.SocketType, endPointWrapper.ProtocolType);
         SetSocketOptions(serverSocket);
 
-        endPoint = endPoint ?? new IPEndPoint(IPAddress.Loopback, 0);
-        serverSocket.Bind(endPoint);
+        if (endPointWrapper.LocalPath != null)
+        {
+          if (File.Exists(endPointWrapper.LocalPath)) File.Delete(endPointWrapper.LocalPath);
+        }
```
ForNeVeR, line 658: A question: shouldn't we expect that the file is delected by the caller? (Not really sure what is the practical answer; I'd just prefer to not delete anything unless really required.)

rd-net/RdFramework/Impl/SocketWire.cs
```diff
@@ -639,25 +643,32 @@ public Server(Lifetime lifetime, IScheduler scheduler, Socket serverSocket, stri
...
 
         return serverSocket;
       }
 
       private void StartServerSocket(Lifetime lifetime, Socket serverSocket)
       {
         if (serverSocket == null) throw new ArgumentNullException(nameof(serverSocket));
-        Port = ((IPEndPoint) serverSocket.LocalEndPoint).Port;
+        Port = serverSocket.LocalEndPoint is IPEndPoint ipEndPoint ? ipEndPoint.Port : null;
+        // TODO do we want to store Path from Unix here? Only via reflection.
```
ForNeVeR, line 670: Perhaps let's keep the `Endpoint` or the `Socket` object somewhere in the public properties, and let the user to sort them out?

rd-net/RdFramework/Tasks/RdFault.cs
```diff
@@ -22,15 +22,17 @@ public RdFault(Exception inner) : base(inner.Message, inner)
...
     protected RdFault(SerializationInfo info, StreamingContext context) : base(info, context)
     {
       ReasonTypeFqn = info.GetString(nameof(ReasonTypeFqn));
       ReasonText = info.GetString(nameof(ReasonText));
       ReasonMessage = info.GetString(nameof(ReasonMessage));
     }
 
-    [SecurityPermission(SecurityAction.Demand, SerializationFormatter = true)]
+    // [SecurityPermission(SecurityAction.Demand, SerializationFormatter = true)]
+    [Obsolete("This API supports obsolete formatter-based serialization. It should not be called or extended by application code.", DiagnosticId = "SYSLIB0051", UrlFormat = "https://aka.ms/dotnet-warnings/{0}")]
```
ForNeVeR, line 35: A note to not forget to do something about these eventually. I'm not sure we use them at all.

rd-net/Test.RdFramework/SocketProxyTest.cs
```diff
@@ -1,127 +1,127 @@
...
-          Assert.AreEqual(new List<int> {1, 2, 4, 5, 6, 8}, clientLog);
-
-          //Connection is broken for now, proxy is not alive
-        }
-      });
-    }
-  }
-}
-#endif
+// #if !NET35
```
ForNeVeR, line 1: Note to uncomment it eventually :)

rd-net/Test.RdFramework/Test.RdFramework.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
         <!--suppress MsbuildTargetFrameworkTagInspection -->
-        <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+        <TargetFramework>net8.0</TargetFramework>
         <TargetFrameworks Condition=" $(OS) == 'Windows_NT' ">net472;$(TargetFrameworks);net35</TargetFrameworks>
```
ForNeVeR, line 5: I think this is broken.

rd-net/Test.Reflection.App/Test.Reflection.App.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net472</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
```
ForNeVeR, line 5: Don't forget to port this to both platforms somehow.

rd-net/global.json
```diff
@@ -1,6 +0,0 @@
-{
-  "sdk": {
-    "version": "6.0.100",
```
ForNeVeR, line 3: Let's update it to net8 better :)

ForNeVeR (changes requested):

rd-net/RdFramework/Impl/SocketWire.cs
ArtManyak: As far as I understood, this is a common recommendation. When you create a UnixSocket you should delete the file if it exists.

rd-net/RdFramework/Impl/SocketWire.cs
ArtManyak: But why? Do we really need this?

PR #504, opened 2025-01-15
Log error on accessing non-initialized lifetime 

Iliya-usov:


ulex:
Have you tried running Rider or ReSharper with this patch applied? I have a feeling that a lot of things will be broken, but I agree this change overall

ForNeVeR (approved):

PR #541, opened 2025-07-25
Fixed infinite loop in UnsafeWriter.Realloc when requesting from 1 to 2 GB of memory!

real-mikhail:
Also, increased the limit of memory that can be allocated from 1 GB to 2 GB (hard limit that current architecture allows).
Also updated script for downloading and setting up .NET for building nugets locally.


ForNeVeR (approved):
Thanks!

PR #559, opened 2025-09-18
add support for net8

Iliya-usov:


rd-net/Lifetimes/Util/BitHacks.cs
```diff
@@ -31,6 +32,19 @@ public static class BitHacks
       13, 18, 8, 12, 7, 6, 5, 63
```
aloraman, line 32: DeBruijin data and `ReverseBitScan()` implementations can be omitted on `NET5_0_OR_GREATER` since their usage is replaced by `System.Numerics.BitOperations` anyway, thus skipping unnecessary array allocations altogether.

ForNeVeR (approved):

PR #606, opened 2026-04-09
Adds SpinOnceWithoutSleep that spins without ever blocking the thread

Iliya-usov:
via sleep. On .NET 5+, delegates to SpinOnce(-1); on older targets, falls back to Thread.Yield() when the next spin would otherwise sleep.

### DIFF
--- a/rd-net/Lifetimes/Threading/SpinWaitEx.cs
+++ b/rd-net/Lifetimes/Threading/SpinWaitEx.cs
@@ -154,5 +154,45 @@ namespace JetBrains.Threading
           return true;
       }
     }
+
+    /// <summary>
+    /// Performs a single spin iteration, guaranteeing that the current thread is never blocked via
+    /// <see cref="Thread.Sleep(int)"/>.
+    /// <para>
+    /// On .NET 5 and later, delegates to <see cref="SpinWait.SpinOnce(int)"/> with
+    /// <c>sleepThreshold = -1</c>, which disables the sleep phase entirely while still allowing
+    /// CPU-level spinning and <see cref="Thread.Yield()"/>.
+    /// </para>
+    /// <para>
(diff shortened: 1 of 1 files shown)


rd-net/Lifetimes/Threading/SpinWaitEx.cs
```diff
@@ -154,5 +154,21 @@ public static async System.Threading.Tasks.Task<bool> SpinUntilAsync(
...
+      spinWait.SpinOnce(-1);
+#else
+      if (spinWait.NextSpinWillYield)
+      {
+        Thread.Yield();
+      }
+      else
+      {
+        spinWait.SpinOnce();
+      }
```
ForNeVeR, line 194: I don't get what you are trying to achieve here.

The base scenario (`NextSpinWillYield == false`) → we call `spinWait.SpinOnce` because we know it won't yield execution to another thread, alright.

The scenario when `NextSpinWillYield == true` → for whatever reason we call `Thread.Yield` ourselves, effectively doing the same as `spinWait.SpinOnce` would?

I have arguments for both variants, but I think  it would be better to document your exact intention here.

ForNeVeR (approved):

rd-net/Lifetimes/Threading/SpinWaitEx.cs
Iliya-usov: The main idea is to avoid `Thread.Sleep(1)` that `SpinOnce()` eventually calls, which can block for ~16 ms on Windows.

ForNeVeR: Thanks, now I get it. `NextSpinWillYield` means that it will call `Sleep(1)`. Makes sense.

ForNeVeR (approved):

PR #613, opened 2026-05-05
Add: FastStopwatch — low-overhead, low-resolution analogue of LocalStopwatch Backed by Environment TickCount on net472/netstandard2.0 and Environment.TickCount64 on .NET 5+.

Iliya-usov:


ulex (approved):

rd-net/Lifetimes/Util/LocalStopwatch.cs
```diff
@@ -56,4 +56,55 @@ private void AssertTimeStamp()
...
+#endif
+
+    public static FastStopwatch StartNew() => new(GetTicks());
+    public long ElapsedMilliseconds
+    {
+      get
+      {
+        AssertTimeStamp();
+        return unchecked(GetTicks() - myStartTimeStamp);
+      }
```
ForNeVeR, line 100: I remember about issues when this was returning negative number in the original `Stopwatch` — is this still an issue? Or is `Environment.TickCount` guaranteed to be monotonic, so this is not a problem?

ForNeVeR (commented):
LGTM.

rd-net/Lifetimes/Util/LocalStopwatch.cs
aloraman: `Environment.TickCount` is monotonic on mainstream systems today. It seems there were some problems with monotonicity on [Mono](https://github.com/mono/mono/pull/7520) and [OSX](https://github.com/dotnet/coreclr/pull/1269), but they are fixed now. There still some inconsistencies between operating systems on whether sleep time is measured by tick count or not - will be changed in [NET11](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/11/environment-tickcount-windows-behavior)

PR #616, opened 2026-05-13
Suppress the compiler warning regarding the javaClass.classLoader

### LINKED ISSUE #492: `javaClass.classLoader` in `registerSerializersCore` causes a compiler warning
Every generated Kotlin class contains this snippet:
```kotlin
        override fun registerSerializersCore(serializers: ISerializers)  {
            val classLoader = javaClass.classLoader
```

`javaClass.classLoader` causes a Kotlin compiler warning:
> The resulting type of this 'javaClass' call is Class<AvaloniaRiderProjectModel.Companion> and not Class<AvaloniaRiderProjectModel>. Please use the more clear '::class.java' syntax to avoid confusion

Let's either replace it with `val classLoader = Companion::class.java.classLoader`, or add `@Suppress("JAVA_CLASS_ON_COMPANION")` to generated files.

aerworker:
Suppress the compiler warning regarding the `javaClass.classLoader` in `registerSerializersCore`
Fix #492


ForNeVeR:
Is there a reason you chose to not use the `Companion::class.java.classLoader` syntax?

aerworker:
There is no specific reason. It's just a minimal change that shouldn't affect behaviour.

PR #618, opened 2026-05-26
Fix RdId colision degradation for legacy identities

ulex:
- Restore old behavior in case of legacy Identities. Use Mix instead of less reliable Next


ForNeVeR:
@Iliya-usov when you get back, I'd like to add some tests for this case.

ForNeVeR (approved):

